PR #47, opened 2024-10-12
fix(rs-lakehouse): creation failing when enable_schemas is false

### LINKED ISSUE #41: [bug] fabric_lakehouse resource not working with Service Principal Secret authentication
### 🐛 What happened?

Currently fabric_lakehouse resource can't be deployed when using Service Principal Client Secret authentication. I'm quite sure the reason is that wrong scopes are used when generating a token for SP to be consumed in the API calls under the hood. 

Currently when trying to deploy, first it gives you an error: 

"Could not create resource: An unexpected error occurred while processing the request"

And if you try second time: 

Could not create resource: Requested '{lakehouse_name}' is already in use

Anyways the lakehouse does not get created and after a while, you can again try with the same name and you end up to the first error. The name is just cached for a while somewhere as reserved.

I have been playing around with Fabric REST APIs and had this same issue when was using wrong scope for token generation. The correct ones to use are:

"https://api.fabric.microsoft.com/Workspace.ReadWrite.All/.default" and "https://api.fabric.microsoft.com/Item.ReadWrite.All/.default"

After using those I'm able to create lakehouses through Service Principal authentication

### 🔬 How to reproduce?

_No response_

### 🏗️ Code Sample / Log

_No response_

### 📷 Screenshots

_No response_

### 📈 Expected behavior

_No response_

0.1.0-beta.4

1.9.7

### 🌌 Environment (OS)

macOS

### 📎 Additional context

_No response_

### 🔰 Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

DariuszPorowski:
# 📥 Pull Request

close #41

This pull request addresses a bug fix in the `fabric_lakehouse` resource creation and enhances the test coverage for different configurations. The key changes include fixing the resource creation logic when schemas are disabled and expanding the test cases to cover more scenarios.

### Bug Fixes:
* [`.changes/unreleased/fixed-20241011-205500.yaml`](diffhunk://#diff-f0c6196f7cc4d5f0d02c62d97da4755c989a3f0741826bc18bd99d816e0a3931R1-R5): Added a new entry to document the fix for the `fabric_lakehouse` resource creation failure when `configuration.enable_schemas` is set to `false`.
* [`internal/services/lakehouse/models_resource_lakehouse.go`](diffhunk://#diff-21525041e68a2b1aa2a83bd03af29dbd976f98b5893b2da2edff25b312968068L38-R38): Corrected the condition to check if `configuration.EnableSchemas` is set to `true` before setting the `CreationPayload`.

### Test Enhancements:
* [`internal/services/lakehouse/resource_lakehouse_test.go`](diffhunk://#diff-788480643cece417ede65d6272edcd8a19d268328028f2712488483c40d61a3eL274-R280): Updated the `TestAcc_LakehouseConfigurationResource_CRUD` test to include additional test cases for creating and updating resources with different `display_name` and `description` values, and both enabled and disabled schemas configurations. [[1]](diffhunk://#diff-788480643cece417ede65d6272edcd8a19d268328028f2712488483c40d61a3eL274-R280) [[2]](diffhunk://#diff-788480643cece417ede65d6272edcd8a19d268328028f2712488483c40d61a3eL286-R297) [[3]](diffhunk://#diff-788480643cece417ede65d6272edcd8a19d268328028f2712488483c40d61a3eL306-R364)

### DIFF
--- a/.changes/unreleased/fixed-20241011-205500.yaml
+++ b/.changes/unreleased/fixed-20241011-205500.yaml
@@ -1,0 +1,5 @@
+kind: fixed
+body: '`fabric_lakehouse` Resource creation failed when `configuration.enable_schemas` has been set to `false`'
+time: 2024-10-11T20:55:00.0241833+02:00
+custom:
+    Issue: "47"
--- a/internal/services/lakehouse/models_resource_lakehouse.go
+++ b/internal/services/lakehouse/models_resource_lakehouse.go
@@ -35,7 +35,7 @@ func (to *requestCreateLakehouse) set(ctx context.Context, from resourceLakehous
 			return diags
 		}
 
-		if !configuration.EnableSchemas.IsNull() && !configuration.EnableSchemas.IsUnknown() {
+		if configuration.EnableSchemas.ValueBool() {
 			to.CreationPayload = &fablakehouse.CreationPayload{
 				EnableSchemas: configuration.EnableSchemas.ValueBoolPointer(),
 			}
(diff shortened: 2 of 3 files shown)


PabloZaiden (approved):
lgtm

PR #140, opened 2024-12-11
refactor(fabricitem): use generics for properties

DariuszPorowski:
# 📥 Pull Request

This pull request includes several changes to the `fabric_spark_job_definition` by adding the extended properties.

Additionally, PR contains internal refactoring for Fabric Items to handle extended properties using Go Generics. This will be the base for refactoring of exiting items to use it and for any future items with properties.


PabloZaiden (approved):
Looks good. Great step forward to generalize a lot of the logic.

Still have some doubts that we might encounter some other resource/datasource that will have issue with the fakes when using the `Filter` method, but so far, that hasn't been an issue. If that causes issues in the future, we might have to do something like we've done for the single entities: looking into the created elements list, and if it's not there, look in the Items element list too.

But good enough for now.

PR #162, opened 2024-12-18
fix(ci): change to src, changes was reserved keyword by action

DariuszPorowski:
# 📥 Pull Request

This pull request updates the `.github/workflows/test.yml` file to change the output variable used for conditional checks from `changes` to `src`.

`changes` keywork is used by action itself and current filter was not working properly - it always run.

bonus: optimized taskfile and test execution

## ✨ Description of new changes

Changes to conditional checks in `.github/workflows/test.yml`


PabloZaiden (approved):
good catch!

Issue #174, opened 2024-12-27
[BUG] Provider produced inconsistent result after apply


Lachlan-White:
### 🐛 What happened?

![Image](https://github.com/user-attachments/assets/5ce55130-7c8e-4791-aa10-3d50ae842cb0)

Error when using a data object to create multiple workspaces for multiple domains at the same time. Using a custom local variable.

### 🔬 How to reproduce?

![Image](https://github.com/user-attachments/assets/cb7336d9-771b-4034-8dea-281726d9408a)

### 📈 Expected behavior

Provisioned Workspaces linked to domains

### 🌌 Environment (Provider Version)

0.1.0-beta.7

### 🌌 Environment (Terraform Version)

1.10.3

### 🌌 Environment (OS)

Windows

### 🔰 Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

DariuszPorowski:
You are trying to assign workspaces to the domain, but you are taking workspace as subject for assignment, so your result is multiple instances of `domain_workspace_assigments`, but must be only 1 per domain - domain is subject, not workspace.

Potential workaround for now:
```terraform
resource "fabric_domain_workspace_assignments" "this" {
  for_each = {
    for domain_key, domain in var.fabric_domains : domain_key => domain
  }
  workspace_ids = [
    for ws_key, ws in local.workspace_configs :
    fabric_workspace.this[ws_key].id if ws.domain_id == try(fabric_domain.parent[each.key].id, fabric_domain.child[each.key].id)
  ]
  domain_id = try(fabric_domain.parent[each.key].id, fabric_domain.child[each.key].id)
}
```

It will be fixed in https://github.com/microsoft/terraform-provider-fabric/pull/182

Lachlan-White:
Perfect @DariuszPorowski workaround is working

PR #182, opened 2025-01-06
fix(domain): workspace assignments should only care about managed resources

DariuszPorowski:
# 📥 Pull Request

- Currently `domain_workspace_assignments` resource enforce assignments based only on TF configuration. It means if someone adds more resources manually on the Fabric side, the next run of TF will remove it. Expected behavior: Resource must only manage TF provided configuration.

This fix is related (not directly) to #174 and potentially to #183

### DIFF
--- a/.changes/unreleased/fixed-20250108-115142.yaml
+++ b/.changes/unreleased/fixed-20250108-115142.yaml
@@ -1,6 +1,6 @@
 kind: fixed
 body: |
-  `Provider produced inconsistent result after apply` error appears when using `fabric_domain_workspace_assignments` multiple times due to enforce assignments based only on TF configuration and ignoring the real state on teh Fabric side.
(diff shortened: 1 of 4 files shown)


.changes/unreleased/fixed-20250108-115142.yaml
```diff
@@ -0,0 +1,7 @@
+kind: fixed
+body: |
+  `Provider produced inconsistent result after apply` error appears when using `fabric_domain_workspace_assignments` multiple times due to enforce assignments based only on TF configuration and ignoring the real state on teh Fabric side.
```
PabloZaiden:
```suggestion
  `Provider produced inconsistent result after apply` error appears when using `fabric_domain_workspace_assignments` multiple times due to enforce assignments based only on TF configuration and ignoring the real state on the Fabric side.
```

PabloZaiden (changes requested):
Looks good. Just suggested a fix for a minor typo

PabloZaiden (approved):

PR #185, opened 2025-01-08
feat!: introduction of preview mode

DariuszPorowski:
# 📥 Pull Request

This pull request includes several changes to the documentation and configuration to introduce a preview mode for various resources and data sources. The changes ensure that users are aware of the risks associated with using preview features and require explicit opt-in at the provider level configuration.


internal/pkg/fabricitem/data_item.go
```diff
@@ -82,6 +83,22 @@ func (d *DataSourceFabricItem) Configure(_ context.Context, req datasource.Confi
 
 	d.pConfigData = pConfigData
 	d.client = fabcore.NewClientFactoryWithClient(*pConfigData.FabricClient).NewItemsClient()
+
+	if d.IsPreview && !d.pConfigData.Preview {
```
PabloZaiden: Is there anything that would prevent us from refactoring this to put this 15-lines block that is repeated several times across the provider in a single place, and just call it from the different components of the provider?

DariuszPorowski: Let me think today about it :)

PR #224, opened 2025-01-28
feat(rs-eventhouse): add configuration support

DariuszPorowski:
# 📥 Pull Request

This pull request introduces several changes to the `fabric_eventhouse` resource, including the addition of a new configuration attribute and updates to the documentation and tests.


examples/resources/fabric_eventhouse/resource.tf
```diff
@@ -9,6 +9,7 @@ resource "fabric_eventhouse" "example_definition_bootstrap" {
   display_name              = "example2"
   description               = "example with definition bootstrapping"
   workspace_id              = "00000000-0000-0000-0000-000000000000"
+  format                    = "Default"
   definition_update_enabled = false # <-- Disable definition update
```
badeamarjieh, line 13: what is this property? `definition_update_enabled`

DariuszPorowski: it's a gatekeeper to determine a behavior when files with definition have changed - update on the Fabric side or not. By default, is `true` but for a certain scenarios can be disabled.

internal/services/eventhouse/resource_eventhouse_test.go
```diff
@@ -111,6 +111,24 @@ func TestUnit_EventhouseResource_Attributes(t *testing.T) {
 				)),
 			ExpectError: regexp.MustCompile(`The argument "display_name" is required, but no definition was found.`),
 		},
+		// error - no required attributes
```
OrBaubergMicrosoft, line 114: maybe a test for creation with config?

Or167 (approved):

internal/services/eventhouse/resource_eventhouse_test.go
DariuszPorowski: it exists (starts at line 394)

PR #237, opened 2025-02-06
feat(logging): enable api/sdk interaction logging

DariuszPorowski:
# 📥 Pull Request

This pull request introduces logging enhancements for API/SDK interactions and updates related documentation and configuration files. The key changes include enabling detailed logging for API calls and updating the troubleshooting guide.

### Logging Enhancements:

* Added a new logging subsystem for the Microsoft Fabric SDK to capture detailed API interactions. (`internal/provider/client/logging.go`)
* Integrated the new logging subsystem into the provider client creation process. (`internal/provider/provider.go`)

### Documentation Updates:

* Updated the troubleshooting guide to include instructions for enabling low-level logging for API calls. (`docs/guides/troubleshooting.md`, `templates/guides/troubleshooting.md`)

### DIFF
--- a/internal/provider/client/logging.go
+++ b/internal/provider/client/logging.go
@@ -17,7 +17,7 @@ func NewFabricSDKLoggerSubsystem(ctx context.Context) (context.Context, hclog.Le
 	targetLevel := hclog.LevelFromString(os.Getenv("FABRIC_SDK_GO_LOGGING"))
 
 	// If the level is not set, or is set to "off", disable logging
-	if targetLevel == hclog.NoLevel || targetLevel == hclog.Off {
+	if targetLevel == hclog.NoLevel {
 		targetLevel = hclog.Off
 	} else {
 		// Enable azcore logging
--- a/internal/provider/client/logging.go
+++ b/internal/provider/client/logging.go
@@ -1,0 +1,34 @@
+// Copyright (c) Microsoft Corporation
+// SPDX-License-Identifier: MPL-2.0
+
+package client
+
+import (
+	"context"
+	"os"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+const FabricSDKLoggerName = "fabric-sdk-go"
+
+func NewFabricSDKLoggerSubsystem(ctx context.Context) (context.Context, hclog.Level, error) {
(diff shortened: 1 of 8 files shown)


internal/provider/client/logging.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+const FabricSDKLoggerName = "fabric-sdk-go"
+
+func NewFabricSDKLoggerSubsystem(ctx context.Context) (context.Context, hclog.Level, error) {
+	targetLevel := hclog.LevelFromString(os.Getenv("FABRIC_SDK_GO_LOGGING"))
+
+	// If the level is not set, or is set to "off", disable logging
+	if targetLevel == hclog.NoLevel || targetLevel == hclog.Off {
```
PabloZaiden: if `targetLevel == hclog.Off`, there's no need to set it again.

```suggestion
	if targetLevel == hclog.NoLevel {
```

internal/provider/provider.go
```diff
@@ -126,9 +129,25 @@ func createDefaultClient(ctx context.Context, cfg *pconfig.ProviderConfig) (*fab
...
 	fabricClientOpt.Retry.MaxRetryDelay = -1
 
+	ctx, lvl, err := pclient.NewFabricSDKLoggerSubsystem(ctx)
+	if err != nil {
+		tflog.Error(ctx, "Failed to initialize Microsoft Fabric SDK logger subsystem", map[string]any{"error": err.Error()})
+
+		return nil, err
+	}
+
+	if cls := os.Getenv("AZURE_SDK_GO_LOGGING"); cls == "all" && lvl != hclog.Off {
```
PabloZaiden: Would it be possible to move `AZURE_SDK_GO_LOGGING` and `all` to reusable constants, and also do case-insensitive comparisons with the values (people might use `ALL` instead of `all` and expect it to work... although now that I think of it, it may always be set from the other part of the code. Anyway, it would be nice to have a constant at least)

PabloZaiden (approved):
Added 2 minor comments, but looks good. Great feature.

internal/provider/client/logging.go
DariuszPorowski: good catch! :)

internal/provider/provider.go
DariuszPorowski: done :)

PR #256, opened 2025-02-14
feat(activator): onboard new resource/data-source

DariuszPorowski:
# 📥 Pull Request

- Add new resource `fabric_activator`
- Add new data-source `fabric_activator`
- Add new data-source `fabric_activators`
- Fix: missing preview note in data-source lists documentation
- Misc: simplifies the code for IsPreviewMode


badeamarjieh:
Activator/Reflex do not support SPN as per the docs. In that case they wouldn't be supported at all here no?

DariuszPorowski:
> Activator/Reflex do not support SPN as per the docs. In that case they wouldn't be supported at all here no?

@badeamarjieh That's the reason we have preview flag here and skip TestAcc tests until Activator team enable SPN support.

badeamarjieh:
@DariuszPorowski  but did we not agree that SPN is mandatory even for resources/data-sources in Terraform Preview Mode?

DariuszPorowski:
@badeamarjieh for GA mandatory

internal/services/activator/data_activator.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/microsoft/terraform-provider-fabric/internal/pkg/fabricitem"
+)
+
+func NewDataSourceActivator() datasource.DataSource {
+	config := fabricitem.DataSourceFabricItemDefinition{
+		Type:   ItemType,
+		Name:   ItemName,
+		TFName: ItemTFName,
+		MarkdownDescription: "Get a Fabric " + ItemName + " (also known as Reflex).\n\n" +
```
PabloZaiden, line 17: Should we put the comment about Reflex, if the official name is Activator?

PabloZaiden (approved):

internal/services/activator/data_activator.go
DariuszPorowski: It was ask from API team.

PR #281, opened 2025-03-05
feat(kql_database): deprecate token and add token_wo

DariuszPorowski:
# 📥 Pull Request

https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments
https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only

### DIFF
--- a/internal/pkg/fabricitem/resource_item_config_properties.go
+++ b/internal/pkg/fabricitem/resource_item_config_properties.go
@@ -79,12 +79,12 @@ func (r *ResourceFabricItemConfigProperties[Ttfprop, Titemprop, Ttfconfig, Titem
 	tflog.Debug(ctx, "CREATE", map[string]any{
 		"action": "start",
 	})
-	tflog.Trace(ctx, "CREATE", map[string]any{
-		"config": req.Config,
-		"plan":   req.Plan,
-	})
 
-	var plan ResourceFabricItemConfigPropertiesModel[Ttfprop, Titemprop, Ttfconfig, Titemconfig]
+	var plan, config ResourceFabricItemConfigPropertiesModel[Ttfprop, Titemprop, Ttfconfig, Titemconfig]
+
+	if resp.Diagnostics.Append(req.Config.Get(ctx, &config)...); resp.Diagnostics.HasError() {
+		return
+	}
 
 	if resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...); resp.Diagnostics.HasError() {
 		return
@@ -104,7 +104,7 @@ func (r *ResourceFabricItemConfigProperties[Ttfprop, Titemprop, Ttfconfig, Titem
 	reqCreate.setDescription(plan.Description)
 	reqCreate.setType(r.Type)
 
-	creationPayload, diags := getCreationPayload(ctx, plan.Configuration, r.CreationPayloadSetter)
+	creationPayload, diags := getCreationPayload(ctx, config.Configuration, r.CreationPayloadSetter)
 	if resp.Diagnostics.Append(diags...); resp.Diagnostics.HasError() {
 		return
 	}
@@ -184,11 +184,6 @@ func (r *ResourceFabricItemConfigProperties[Ttfprop, Titemprop, Ttfconfig, Titem
 	tflog.Debug(ctx, "UPDATE", map[string]any{
 		"action": "start",
 	})
-	tflog.Trace(ctx, "UPDATE", map[string]any{
-		"config": req.Config,
-		"plan":   req.Plan,
-		"state":  req.State,
-	})
 
 	var plan, state ResourceFabricItemConfigPropertiesModel[Ttfprop, Titemprop, Ttfconfig, Titemconfig]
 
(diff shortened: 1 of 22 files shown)


internal/pkg/fabricitem/resource_item_config_properties.go
```diff
@@ -104,7 +104,7 @@ func (r *ResourceFabricItemConfigProperties[Ttfprop, Titemprop, Ttfconfig, Titem
 	reqCreate.setDescription(plan.Description)
 	reqCreate.setType(r.Type)
 
-	creationPayload, diags := getCreationPayload(ctx, plan.Configuration, r.CreationPayloadSetter)
+	creationPayload, diags := getCreationPayload(ctx, config.Configuration, r.CreationPayloadSetter)
```
OrBaubergMicrosoft, line 107: getCreationPayload takes Tconfiguration as an argument, how did it work without it?
because the property setter uses the type specific configuration properties , and they exist also in the plan?

internal/services/kqldatabase/models_resource_kql_database.go
```diff
@@ -10,9 +10,11 @@ import (
...
-	DatabaseType       types.String     `tfsdk:"database_type"`
-	EventhouseID       customtypes.UUID `tfsdk:"eventhouse_id"`
-	InvitationToken    types.String     `tfsdk:"invitation_token"`
-	SourceClusterURI   customtypes.URL  `tfsdk:"source_cluster_uri"`
-	SourceDatabaseName types.String     `tfsdk:"source_database_name"`
+	DatabaseType             types.String     `tfsdk:"database_type"`
+	EventhouseID             customtypes.UUID `tfsdk:"eventhouse_id"`
+	InvitationToken          types.String     `tfsdk:"invitation_token"`
+	InvitationTokenWO        types.String     `tfsdk:"invitation_token_wo"`
+	InvitationTokenWOVersion types.Int32      `tfsdk:"invitation_token_wo_version"`
```
OrBaubergMicrosoft, line 17: the invitation token is a part of the creation payload, meaning that currently it cannot be update.  why is the versioning needed?
I think it's only relevant for read only & updateable properties.

internal/services/kqldatabase/schema_resource_kql_database.go
```diff
@@ -69,18 +71,57 @@ func getResourceKQLDatabaseConfigurationAttributes() map[string]schema.Attribute
 			MarkdownDescription: "Invitation token to follow the source database. Only allowed when `database_type` is `" + string(fabkqldatabase.TypeShortcut) + "`.",
```
OrBaubergMicrosoft, line 71: how do we plan to depricate it?
will we write deprication date? 
do we have any traces for invitation_token usage vs _wo usage?

internal/pkg/fabricitem/resource_item_config_properties.go
DariuszPorowski: @OrBaubergMicrosoft Tconfiguration is not argument. It's a type indicator. getCreationPayload is aware of type because function receiver has generic implementation with types used by getCreationPayload.

internal/services/kqldatabase/models_resource_kql_database.go
DariuszPorowski: @OrBaubergMicrosoft change of token triggers recreate operation, and because "WO" not trigger any change, additional "version" is a keeper. ref: https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments#best-practices

internal/services/kqldatabase/schema_resource_kql_database.go
DariuszPorowski: @OrBaubergMicrosoft No, we are not tracking it. It will be removed on the next major release based on guidelines here: https://developer.hashicorp.com/terraform/plugin/framework/deprecations
for now, we are keeping both ways, old and new.

internal/services/kqldatabase/schema_resource_kql_database.go
```diff
@@ -69,18 +71,57 @@ func getResourceKQLDatabaseConfigurationAttributes() map[string]schema.Attribute
...
+			DeprecationMessage:  "This attribute is deprecated. Use `invitation_token_wo` instead.",
 			PlanModifiers: []planmodifier.String{
 				stringplanmodifier.RequiresReplace(),
 			},
+			Validators: []validator.String{
+				stringvalidator.PreferWriteOnlyAttribute(
+					path.MatchRelative().AtParent().AtName("invitation_token_wo"),
+				),
+				stringvalidator.ConflictsWith(
+					path.MatchRelative().AtParent().AtName("invitation_token_wo"),
```
badeamarjieh, line 83: maybe add conflicting with `invitation_token_wo_version` as well

badeamarjieh (commented):
please update the exmaples to use _wo. and probably add an example to updating the wo version?

internal/services/kqldatabase/schema_resource_kql_database.go
DariuszPorowski: makes sense

badeamarjieh (approved):

PR #286, opened 2025-03-05
feat(domain): enable spn support

DariuszPorowski:
# 📥 Pull Request

This pull request introduces support for Service Principal (SPN) authentication for various domain-related resources and data sources. The changes span across documentation updates, configuration files, and test cases.

### DIFF
--- a/internal/services/domain/resource_domain_role_assignments.go
+++ b/internal/services/domain/resource_domain_role_assignments.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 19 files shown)


internal/services/domain/resource_domain_role_assignments.go
```diff
@@ -335,10 +332,30 @@ func (r *resourceDomainRoleAssignments) Delete(ctx context.Context, req resource
 	}
 
 	_, err := r.client.RoleAssignmentsBulkUnassign(ctx, state.DomainID.ValueString(), reqDelete.DomainRoleUnassignmentRequest, nil)
-	if resp.Diagnostics.Append(utils.GetDiagsFromError(ctx, err, utils.OperationDelete, nil)...); resp.Diagnostics.HasError() {
+	diags = utils.GetDiagsFromError(ctx, err, utils.OperationDelete, fabcore.ErrDomain.DomainSpecificUsersScopeCannotBeEmptyError)
```
PabloZaiden, line 335: I get why we're doing this here, but it feels a bit like deciding the flow of execution based on the type of errors we receive (like doing try-catch with specific type of exception in some other languages.) Are you sure we cannot infer it would happen, based on the current assignments and scope and the updated values, instead of trying and failing first?

PabloZaiden (approved):

internal/services/domain/resource_domain_role_assignments.go
DariuszPorowski: It may be the case that a few roles are managed outside of Terraform, and during the deletion operation on our end, Terraform removes only managed roles and leaves the rest - this is expected behavior.
`DomainSpecificUsersScopeCannotBeEmptyError` only occurs when all managed roles on the Terraform side equal roles on the services side. When this happens, the API returns this error, indicating that it is safe to use this workaround without failing.

PR #297, opened 2025-03-11
feat(ds-workspace_role_assignments)!: update principal schema

DariuszPorowski:
# 📥 Pull Request

## ❓ What are you trying to address

This pull request introduces updates the principal schema for `fabric_workspace_role_assignments` - naming unification with all resources/data-sources (breaking).

PR #316, opened 2025-03-16
refactor(domain): use superschema

DariuszPorowski:
# 📥 Pull Request

SuperSchema PoC - domain has been picked up as a good initial example.

SuperSchema allows you to write the Terraform schema for resources and data-sources in a single definition and solves a few common problems by:

- Avoid code duplication on common fields applied on resources and data-sources
- Auto-format attributes like markdown description with validators, plan modifiers, descriptions, default values... - increased value in the auto-generated documentation
  - Enum values auto-injected to the docs, instead of handling it in the code - cleaner code after all.
- Enforce "resource - data-source" schema consistency; example bug what we had: https://github.com/microsoft/terraform-provider-fabric/pull/297
- Keep single, consolidated source of truth for schemas - increased developer cognitive perception of the schema.


internal/services/domain/schema_domain.go
```diff
@@ -0,0 +1,145 @@
...
+func domainSchema(dsList bool) superschema.Schema { //revive:disable-line:flag-parameter
+	markdownDescriptionR := "The " + ItemName + " resource allows you to manage [" + ItemName + "](" + ItemDocsURL + ")."
+	markdownDescriptionR = fabricitem.GetResourceSPNSupportNote(markdownDescriptionR, ItemSPNSupport)
+	markdownDescriptionR = fabricitem.GetResourcePreviewNote(markdownDescriptionR, ItemPreview)
+
+	var dsTimeout *superschema.DatasourceTimeoutAttribute
+	var markdownDescriptionD string
+
+	if dsList {
+		markdownDescriptionD = "The " + ItemName + " data-source allows you to read a list of [" + ItemName + "](" + ItemDocsURL + ") details."
```
badeamarjieh, line 32: could we have a similar method to `GetResourceSPNSupportNote` for the markdown description of DS/DSs/RS?

badeamarjieh: or even maybe combine the three methods of `GetSPNSupportNote`, `GetPreviewNote` and the suggested function to create this text into one method `GetMarkdownDescription` which takes as parameters all it needs to create the description, e.g.,  ItemName, ItemDocsURL, ItemSPNSupport, ItemPreview, isResource/isDS.

internal/services/domain/schema_domain.go
```diff
@@ -0,0 +1,145 @@
...
+				},
+			},
+			"timeouts": superschema.TimeoutAttribute{
+				Resource: &superschema.ResourceTimeoutAttribute{
+					Create: true,
+					Read:   true,
+					Update: true,
+					Delete: true,
+				},
+				DataSource: dsTimeout,
```
badeamarjieh, line 142: when `dsList` is true, we don't want this "timeouts" property at all as it should be in the outer layer. in this case `dsTimeout` is *superschema.DatasourceTimeoutAttribute, does it achieve this?

internal/services/domain/schema_domain.go
```diff
@@ -0,0 +1,145 @@
...
+					Validators: []validator.String{
+						stringvalidator.OneOf(utils.ConvertEnumsToStringSlices(fabadmin.PossibleContributorsScopeTypeValues(), true)...),
+					},
+				},
+				Resource: &schemaR.StringAttribute{
+					Optional: true,
+					Computed: true,
+					Validators: []validator.String{
+						stringvalidator.ConflictsWith(path.MatchRoot("parent_domain_id")),
+					},
```
badeamarjieh, line 128: where is this coming from?

internal/services/domain/schema_domain.go
```diff
@@ -0,0 +1,145 @@
...
+				DataSource: &schemaD.StringAttribute{
+					Computed: true,
+				},
+			},
+			"contributors_scope": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "The " + ItemName + " contributors scope.",
+					Validators: []validator.String{
+						stringvalidator.OneOf(utils.ConvertEnumsToStringSlices(fabadmin.PossibleContributorsScopeTypeValues(), true)...),
+					},
```
badeamarjieh, line 120: this should be in `Resource` instead.

internal/services/domain/schema_domain.go
```diff
@@ -0,0 +1,145 @@
...
+						stringplanmodifier.RequiresReplace(),
+					},
+				},
+				DataSource: &schemaD.StringAttribute{
+					Computed: true,
+				},
+			},
+			"contributors_scope": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "The " + ItemName + " contributors scope.",
```
badeamarjieh, line 118: add `Possible values: " + utils.ConvertStringSlicesToString(fabadmin.PossibleContributorsScopeTypeValues(), true, true) + ".",` ?

internal/services/domain/schema_domain.go
```diff
@@ -0,0 +1,145 @@
...
+					Validators: []validator.String{
```
badeamarjieh: the current md description is slighltly different between the resource and data-source.
I guess its best if we remove it from Common and define it differently for Resource and Datasource.

internal/services/domain/schema_domain.go
badeamarjieh: is this the techincal translation of the existing "Contributors scope can only be set at the root domain level."?

internal/services/domain/schema_domain.go
DariuszPorowski: @badeamarjieh makes sense. Let's put this in a separate PR that will be dedicated only for MarkdownDescription and related refactoring.

internal/services/domain/schema_domain.go
DariuszPorowski: indeed!

internal/services/domain/schema_domain.go
DariuszPorowski: No, its expected to have in common. For data-source its a source for docs, without real validation when attribute is computed.

internal/services/domain/schema_domain.go
DariuszPorowski: good catch

internal/services/domain/schema_domain.go
DariuszPorowski: yes, because its nil

internal/services/domain/schema_domain.go
badeamarjieh: oh nice

PR #324, opened 2025-03-19
feat(spark_workspace_settings): add option to manage default pool using id

DariuszPorowski:
# 📥 Pull Request

This pull request introduces an option to manage the "default pool" using the pool ID in the `fabric_spark_workspace_settings` resource. The changes span documentation updates, schema modifications, and test enhancements.

### DIFF
--- a/.changes/unreleased/added-20250318-195402.yaml
+++ b/.changes/unreleased/added-20250318-195402.yaml
@@ -1,0 +1,5 @@
+kind: added
+body: Option to manage "default pool" using pool ID in the `fabric_spark_workspace_settings` resource.
+time: 2025-03-18T19:54:02.3978618-07:00
+custom:
+    Issue: "111"
--- a/.changes/unreleased/added-20250318-195402.yaml
+++ b/.changes/unreleased/added-20250318-195402.yaml
@@ -2,4 +2,4 @@ kind: added
 body: Option to manage "default pool" using pool ID in the `fabric_spark_workspace_settings` resource.
 time: 2025-03-18T19:54:02.3978618-07:00
 custom:
-    Issue: "111"
+    Issue: "324"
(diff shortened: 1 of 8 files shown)


.changes/unreleased/added-20250318-195402.yaml
```diff
@@ -0,0 +1,5 @@
+kind: added
+body: Option to manage "default pool" using pool ID in the `fabric_spark_workspace_settings` resource.
+time: 2025-03-18T19:54:02.3978618-07:00
+custom:
+    Issue: "111"
```
badeamarjieh: seems like a different bug

internal/services/spark/resource_spark_workspace_settings.go
```diff
@@ -206,22 +206,30 @@ func (r *resourceSparkWorkspaceSettings) Schema(ctx context.Context, _ resource.
...
 								Computed:            true,
 								CustomType:          customtypes.UUIDType{},
 								PlanModifiers: []planmodifier.String{
 									stringplanmodifier.UseStateForUnknown(),
 								},
+								Validators: []validator.String{
+									stringvalidator.ConflictsWith(
+										path.MatchRelative().AtParent().AtName("name"),
+										path.MatchRelative().AtParent().AtName("type"),
+									),
```
badeamarjieh, line 220: is name unique per type only? also, I don't see in the API documentation that id conflicts with these two properties. is this a documentation gap?
https://learn.microsoft.com/en-us/rest/api/fabric/spark/workspace-settings/update-spark-settings?tabs=HTTP#instancepool

DariuszPorowski: It looks like its missing in public docs.

badeamarjieh (approved):

badeamarjieh (approved):

Issue #422, opened 2025-05-05
[RS] Add onelake shortcut resource


catalina-gav:
### 📝 Description

When I am automating and managing data analytics infrastructure using Microsoft Fabric, I want to provision and configure folders as code in Terraform.
To consistently manage folders as code in Terraform while automating and managing data analytics infrastructure using Microsoft Fabric.

### 🔬 Details / References

- Resource Name: `fabric_onelake_shortcut`
- API documentation: 
  -  Create https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/create-shortcut?tabs=HTTP#target
  - Delete https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/delete-shortcut?tabs=HTTP
  - Get https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/get-shortcut?tabs=HTTP
  - List https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/list-shortcuts?tabs=HTTP
- Estimated complexity/effort: moderate
- Related resources/data-sources: https://learn.microsoft.com/en-us/fabric/onelake/onelake-shortcuts

### 🚧 Potential Terraform Configuration / Desired Solution

```terraform
resource "fabric_onelake_shortcut" "onelake" {
  workspace_id = "00000000-0000-0000-0000-000000000000"
  item_id      = "00000000-0000-0000-0000-000000000000"
  name         = "MyShortcutName"
  path         = "/MyShortcutPath"
  target = {
    onelake = {
      workspace_id = "00000000-0000-0000-0000-000000000000"
      item_id      = "00000000-0000-0000-0000-000000000000"
      path         = "/MyTargetPath"
    }
  }
}

resource "fabric_onelake_shortcut" "adlsgen2" {
  workspace_id = "00000000-0000-0000-0000-000000000000"
  item_id      = "00000000-0000-0000-0000-000000000000"
  name         = "MyShortcutName"
  path         = "/MyShortcutPath"
  target = {
    adlsGen2 = {
      location      = "MyLocation"
      subpath       = "MySubpath"
      connection_id = "00000000-0000-0000-0000-000000000000"
    }
  }
}

resource "fabric_onelake_shortcut" "amazon_s3" {
  workspace_id = "00000000-0000-0000-0000-000000000000"
  item_id      = "00000000-0000-0000-0000-000000000000"
[...]

Issue #421, opened 2025-05-05
[DS] Add onelake shortcut data source


catalina-gav:
### 📝 Description

When I am automating and managing data analytics infrastructure using Microsoft Fabric, I want to get details of onelake shortcuts as code in Terraform.
To consistently manage onelake shortcuts as code in Terraform while automating and managing data analytics infrastructure using Microsoft Fabric.

### 🔬 Details / References

- Data Source Name: `fabric_onelake_shortcut`
- API documentation: 
  -  Create https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/create-shortcut?tabs=HTTP#target
  - Delete https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/delete-shortcut?tabs=HTTP
  - Get https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/get-shortcut?tabs=HTTP
  - List https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/list-shortcuts?tabs=HTTP
- Estimated complexity/effort: moderate
- Related resources/data-sources: https://learn.microsoft.com/en-us/fabric/onelake/onelake-shortcuts

### 🚧 Potential Terraform Configuration / Desired Solution

```terraform
data "fabric_onelake_shortcut" "example_by_name_path_wsid_itemid" {
  name   = "MyShortcutName"
  path   = "/MyShortcutPath"
  wsid   = "00000000-0000-0000-0000-000000000000"
  itemid = "00000000-0000-0000-0000-000000000000"
}
data "fabric_onelake_shortcuts" "example" {
  workspace_id = "00000000-0000-0000-0000-000000000000"
  item_id      = "00000000-0000-0000-0000-000000000000"
}

```

### ✅ Definition of Done

- [ ] Data Transfer Objects (DTOs)
- [ ] Data-Source Implementation
- [ ] Data-Source Added to Provider
- [ ] Unit Tests for Happy path
- [ ] Unit Tests for Error path
- [ ] Acceptance Tests
- [ ] Example in the ./examples folder
- [ ] Schema documentation in code
- [ ] Updated auto-generated provider docs with `task docs`

### 🔰 Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

PR #424, opened 2025-05-06
feat(onelake_shortcut): onboard new data-source and resource

catalina-gav:
# 📥 Pull Request

close https://github.com/microsoft/terraform-provider-fabric/issues/422
close https://github.com/microsoft/terraform-provider-fabric/issues/421

Onboard new onelake shortcut resource/data-source


internal/services/onelakeshortcut/schema.go
```diff
@@ -0,0 +1,664 @@
...
+					PlanModifiers: []planmodifier.String{
+						stringplanmodifier.RequiresReplace(),
+					},
+				},
+				DataSource: &schemaD.StringAttribute{
+					Optional: !isList,
+					Computed: true,
+				},
+			},
+			"target": superschema.SuperSingleNestedAttributeOf[targetModel]{
```
DariuszPorowski: `target.type` attribute is missing, that will help end-users to handle their own conditions on the module-level.

internal/services/onelakeshortcut/schema.go
```diff
@@ -0,0 +1,664 @@
...
+								},
+								DataSource: &schemaD.StringAttribute{
+									Computed: true,
+								},
+							},
+						},
+					},
+					"adls_gen2": superschema.SuperSingleNestedAttributeOf[adlsGen2]{
+						Common: &schemaR.SingleNestedAttribute{
+							Optional:            true,
```
DariuszPorowski: `Optional` is not applicable for data-source in this scenario (the same feedback for each of the target types).

internal/services/onelakeshortcut/schema.go
DariuszPorowski: For the resource, the validator is missing, that will check only one-type of shortcut is possible at time, and any change to type will triggers recreate.

internal/services/onelakeshortcut/schema.go
```diff
@@ -0,0 +1,664 @@
...
+					},
+				},
+				DataSource: &schemaD.StringAttribute{
+					Required: !isList,
+					Computed: isList,
+				},
+			},
+			"path": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "A string representing the full path where the shortcut is created, including either \"Files\" or \"Tables\".",
```
DariuszPorowski: nit - the syntax can be more readable.

```suggestion
					MarkdownDescription: `A string representing the full path where the shortcut is created, including either "Files" or "Tables".`,
```

DariuszPorowski (changes requested):
A few early stage feedback comments

examples/data-sources/fabric_onelake_shortcut/data-source.tf
```diff
@@ -0,0 +1,6 @@
+data "fabric_onelake_shortcut" "example_by_name_path_wsid_itemid" {
+  name   = "MyShortcutName"
+  path   = "/MyShortcutPath"
+  wsid   = "00000000-0000-0000-0000-000000000000"
```
DariuszPorowski: workspace_id?

examples/data-sources/fabric_onelake_shortcut/data-source.tf
```diff
@@ -0,0 +1,6 @@
...
+  itemid = "00000000-0000-0000-0000-000000000000"
```
DariuszPorowski: item_id?

internal/services/onelakeshortcut/schema.go
catalina-gav: I added the target.type, but it will always be computed as it it not used upon creation in the API as well (it  can be sent in the body, but has no validation: can send type adlsGen2 and the target to be an onelake and it will work).

catalina-gav: It is being validated in the onelake_shortcut_resource.go as I configured it as a ResourceWithValidateConfig

catalina-gav: It is being validated in the onelake_shortcut_resource.go as I configured it as a ResourceWithConfigValidators using resourceValidator, I am not using type for validation since it's not validated in the API either and not used for create/update. 
And from what I've asked the API team, changing the target from a source to another ( eq onelake to amazon_s3) will update the existing item, not create a new one. So I configured it to be updated in the provider as well, instead of triggering recreate.

catalina-gav: @DariuszPorowski

catalina-gav:
There are 2 fake servers configurations in this pr draft as of now, and should be decided which one should be kept:

**First implementation:**

- implemented fabric_onelakeshortcut following patterns used for other items.
- Needed to create a new configureOneLakeShortcutHandler to match the core SDK's method signatures for OneLake shortcut (e.g., Get requires 4 parameters: workspace ID, item ID, path, and name). This involved creating a new operation and implementing new handle methods for the configuration
- The actual provider composes the ID using workspaceID/itemID/path/name, so there’s a slight inconsistency between handler logic and provider logic, because in order to not duplicate code and create a new typeHandler with an identifier taking 4 parameters, I used the GetId from the existing identifier where I cant access the workspaceId and ItemId URI parameters, so only composed it as path/name

**Second implementation:**

- Follows the examples from items like workspace git which have the fake functions implemented separately in a fake_test file under the same item folder. 
- Implements explicit fake functions for each CRUD operation: fakeCreateOneLakeShortcutFunc, fakeGetOneLakeShortcutFunc, etc., handling the cases and returns for item not found, conflicts etc. 
- Uses a `var fakeOneLakeShortcutStore = map[string]fabcore.Shortcut{}`  in-memory store used to keep and identify data sources and resource in the fake server: if we simulate a create using fakeCreateOneLakeShortcutFunc(), we add a record to the map, if we want to retrieve it, we look up the shortcut in the map. 
- Also upon entity creation, the entity is added to the store, same manner as we would have used handler.Upsert for the other method.

OrBaubergMicrosoft:
> There are 2 fake servers configurations in this pr draft as of now, and should be decided which one should be kept:
> 
> **First implementation:**
> 
> * implemented fabric_onelakeshortcut following patterns used for other items.
> * Needed to create a new configureOneLakeShortcutHandler to match the core SDK's method signatures for OneLake shortcut (e.g., Get requires 4 parameters: workspace ID, item ID, path, and name). This involved creating a new operation and implementing new handle methods for the configuration
> * The actual provider composes the ID using workspaceID/itemID/path/name, so there’s a slight inconsistency between handler logic and provider logic, because in order to not duplicate code and create a new typeHandler with an identifier taking 4 parameters, I used the GetId from the existing identifier where I cant access the workspaceId and ItemId URI parameters, so only composed it as path/name
> 
> **Second implementation:**
> 
> * Follows the examples from items like workspace git which have the fake functions implemented separately in a fake_test file under the same item folder.
> * Implements explicit fake functions for each CRUD operation: fakeCreateOneLakeShortcutFunc, fakeGetOneLakeShortcutFunc, etc., handling the cases and returns for item not found, conflicts etc.
> * Uses a `var fakeOneLakeShortcutStore = map[string]fabcore.Shortcut{}`  in-memory store used to keep and identify data sources and resource in the fake server: if we simulate a create using fakeCreateOneLakeShortcutFunc(), we add a record to the map, if we want to retrieve it, we look up the shortcut in the map.
> * Also upon entity creation, the entity is added to the store, same manner as we would have used handler.Upsert for the other method.

internal/services/onelakeshortcut/resource_onelake_shortcut.go
```diff
@@ -0,0 +1,319 @@
...
+
+	state.Timeouts = plan.Timeouts
+
+	var reqCreate requestCreateOnelakeShortcut
+
+	if resp.Diagnostics.Append(reqCreate.set(ctx, plan)...); resp.Diagnostics.HasError() {
+		return
+	}
+
+	overwriteOnlyPolicy := fabcore.ShortcutConflictPolicyOverwriteOnly
```
DariuszPorowski: To discuss

OrBaubergMicrosoft: discussion decision, ShortcutConflictPolicy will be managed implicitly, a reddit post will be sent on matter, and we will wait for user's perspectives

tools/scripts/Set-WellKnown.ps1
```diff
@@ -800,6 +800,35 @@ function Set-AzureDataFactory {
...
+
+  # Attempt to get the existing shortcut
+  $results = Invoke-FabricRest -Method 'GET' -Endpoint "workspaces/$WorkspaceId/items/$ItemId/shortcuts"
+  $result = $results.Response.value | Where-Object { $_.name -eq $Payload.name -and ($_.path.TrimStart('/') -eq $Payload.path.TrimStart('/')) }
+
+  if (!$result) {
+    # Shortcut does not exist, so create it
+    Write-Log -Message "Creating Shortcut: $($Payload.name)" -Level 'INFO'
+
+    $result = (Invoke-FabricRest -Method 'POST' -Endpoint "workspaces/$WorkspaceId/items/$ItemId/shortcuts" -Payload $Payload).Response
```
OrBaubergMicrosoft: please remove

internal/services/shortcut/base.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/microsoft/terraform-provider-fabric/internal/pkg/tftypeinfo"
+)
+
+var ItemTypeInfo = tftypeinfo.TFTypeInfo{ //nolint:gochecknoglobals
+	Name:           "Shortcut",
+	Type:           "shortcut",
+	Names:          "Shortcuts",
+	Types:          "shortcuts",
+	DocsURL:        "https://learn.microsoft.com/fabric/onelake/onelake-shortcuts",
+	IsPreview:      false,
```
OrBaubergMicrosoft: @DariuszPorowski @badeamarjieh 
IMO this should be in preview mode, and we will turn it to GA when we decide on explicit or implicit shortcut policy handling
WDYT?

internal/services/shortcut/schema.go
```diff
@@ -0,0 +1,578 @@
...
+							MarkdownDescription: "The type object contains properties like target shortcut account type. Additional types may be added over time.",
+						},
+						Resource: &schemaR.StringAttribute{
+							Computed: true,
+						},
+						DataSource: &schemaD.StringAttribute{
+							Computed: true,
+						},
+					},
+					"onelake":              onelakeSchema(),
```
OrBaubergMicrosoft, line 140: can you move validations that only one is populated to here?

internal/services/shortcut/base.go
DariuszPorowski: Lets mark as preview...

internal/services/shortcut/schema.go
catalina-gav: I'm trying to move a validation rule from the ConfigValidators section in resource_shortcut.go into the schema definition using objectvalidator.ExactlyOneOf. Specifically, I’ve replaced the original resourcevalidator.ExactlyOneOf(...) with this schema-level code:
```
Validators: []validator.Object{
    objectvalidator.ExactlyOneOf(
        path.MatchRelative().AtName("onelake"), //tried with matchRoot as it was in ConfigValidators too
        path.MatchRelative().AtName("adls_gen2"),
        path.MatchRelative().AtName("amazon_s3"),
        ...
    ),
},
```
I’ve seen similar usage with validators like RequireIfAttributeIsOneOf and NullIfAttributeIsOneOf, which work correctly inside the schema. However, after switching to objectvalidator.ExactlyOneOf, I'm seeing unexpected behavior during unit tests: the cases that previously passed using resourcevalidator.ExactlyOneOf now fail with the new schema validation.

I was wondering if we can even move this validation ConfigValidators to schema-level, and if so, any guidance on the correct usage or how to structure this validation inside the schema would be much appreciated. Thanks!

catalina-gav: We'll keep the target validation under ConfigValidator, solving this comment

Issue #429, opened 2025-05-12
[RS-PROP] Enhance SQL Database item resource by adding properties support 


catalina-gav:
### 📝 Description

When I am managing my Infrastructure as Code using Terraform for SQL Database. I want to be able to get properties details. So, I can use it as input to another resource/data-source.

### 🔬 Details / References

- Resource Name: `fabric_sql_database`
- API documentation: 
  - GET https://learn.microsoft.com/en-us/rest/api/fabric/sqldatabase/items/get-sql-database?tabs=HTTP
  - LIST https://learn.microsoft.com/en-us/rest/api/fabric/sqldatabase/items/list-sql-databases?tabs=HTTP
- Estimated complexity/effort: easy
- Related resources/data-sources: https://learn.microsoft.com/en-us/fabric/database/sql/overview

### 🚧 Potential Terraform Configuration / Desired Solution

```terraform
# output pseudocode
{
  # existing attributes
  # ...

  properties = {
    connection_string = "MyConnectionString",
    database_name = "MyDatabaseName",
    server_fqdn = "MyServerFqdn"
  }
}
```

### ✅ Definition of Done

- [ ] Data Transfer Objects (DTOs)
- [ ] Resource Implementation
- [ ] Resource Added to Provider
- [ ] Unit Tests for Happy path
- [ ] Unit Tests for Error path
- [ ] Acceptance Tests
- [ ] Example in the ./examples folder
- [ ] Schema documentation in code
- [ ] Updated auto-generated provider docs with `task docs`

### 🔰 Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

PR #430, opened 2025-05-12
feat(sql_database): enhance the SQL database item by adding properties support

catalina-gav:
# 📥 Pull Request
close https://github.com/microsoft/terraform-provider-fabric/issues/429
close https://github.com/microsoft/terraform-provider-fabric/issues/428
Properties have been added to the SQL Database item: https://learn.microsoft.com/en-us/rest/api/fabric/sqldatabase/items/get-sql-database?tabs=HTTP#sqldatabaseproperties


internal/services/sqldatabase/schema_resource_sql_database.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation
+// SPDX-License-Identifier: MPL-2.0
+
+package sqldatabase
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+)
+
+func getResourceSQLDatabasePropertiesAttributes() map[string]schema.Attribute {
```
OrBaubergMicrosoft, line 10: can we use super schema instead?

DariuszPorowski: "fabric item" generic implementation does not handle superschema today.

DariuszPorowski (approved):
LGTM

PR #501, opened 2025-06-14
feat(item): enhance update item to handle ItemDisplayNameNotAvailableYet when display name is reserved for 5 min 

badeamarjieh:
# 📥 Pull Request

- Upon Create or Update Item, item display name could be reserved for up to 5 minutes. In this case, the http call returns with error code `ItemDisplayNameNotAvailableYet`
- This PR handles the above case by applying a retry update mechanism.

### DIFF
--- a/internal/services/lakehouse/resource_lakehouse_test.go
+++ b/internal/services/lakehouse/resource_lakehouse_test.go
@@ -285,7 +285,6 @@ func TestAcc_LakehouseResource_CRUD(t *testing.T) {
 			),
 			Check: resource.ComposeAggregateTestCheckFunc(
 				resource.TestCheckResourceAttr(testResourceItemFQN, "display_name", entityUpdateDisplayName),
-				resource.TestCheckNoResourceAttr(testResourceItemFQN, "configuration"),
 				resource.TestCheckResourceAttr(testResourceItemFQN, "configuration.enable_schemas", "true"),
 			),
 		},


badeamarjieh:
add tests

internal/pkg/fabricitem/helpers.go
```diff
@@ -130,3 +135,60 @@ func IsPreviewMode(name string, itemIsPreview, providerPreviewMode bool) diag.Di
...
+	var errRespFabric *fabcore.ResponseError
+
+	for i := 0; i < config.MaxRetries; i++ {
+		err = operation()
+
+		if err == nil {
+			return nil
+		}
+
+		if errors.As(err, &errRespFabric) && errRespFabric.ErrorCode == "ItemDisplayNameNotAvailableYet" {
```
badeamarjieh: use ErrCommon once this err code is populated

internal/pkg/fabricitem/helpers.go
```diff
@@ -130,3 +135,94 @@ func IsPreviewMode(name string, itemIsPreview, providerPreviewMode bool) diag.Di
...
+		RetryInterval: 2 * time.Minute,
+		Operation:     "update",
+	}
+}
+
+func RetryUpdateItem(ctx context.Context, client *fabcore.ItemsClient, workspaceID, itemID string, request fabcore.UpdateItemRequest) (fabcore.ItemsClientUpdateItemResponse, error) {
+	var respUpdate fabcore.ItemsClientUpdateItemResponse
+	var err error
+
+	err = RetryOperation(ctx, DefaultUpdateRetryConfig(), func() error {
```
OrBaubergMicrosoft: IMO updating inside of the retry operation is a bit odd,
why not pulling the update to be done before the retry function call, and enter it only if there is an error

badeamarjieh: Inside the retry function I'd need to call update again, so that's duplicate code.

badeamarjieh: alternatively, I can call the outer function "CreateItem" or "UpdateItem" and the internal implementation can reflect the retry mechanism, WDYT?

main.tf
```diff
@@ -0,0 +1,29 @@
...
+terraform {
+  required_version = ">= 1.8, < 2.0"
+  required_providers {
+    fabric = {
+      source  = "microsoft/fabric"
+      version = "1.2.0"
+    }
+  }
+}
+
```
OrBaubergMicrosoft: revert (?)

internal/pkg/fabricitem/helpers.go
OrBaubergMicrosoft: > Inside the retry function I'd need to call update again, so that's duplicate code.

why?

func UpdateItem(ctx context.Context, client *fabcore.ItemsClient, workspaceID, itemID string, request fabcore.UpdateItemRequest) (fabcore.ItemsClientUpdateItemResponse, error) {
	var respUpdate fabcore.ItemsClientUpdateItemResponse
	var shouldRetry = True
	while (shouldRetry)
	{
            respUpdate, err = client.UpdateItem(ctx, workspaceID, itemID, request, nil)
            if err != nil && error
            {
	        shouldRetry =RetryOperation(ctx, DefaultUpdateRetryConfig(), error
	        }
	    

inside to retry check if it is the right error code

PR #522, opened 2025-06-30
feat(docs): update FAQ

DariuszPorowski:


templates/guides/troubleshooting.md
```diff
@@ -148,11 +148,12 @@ terraform apply
 
 ## FAQ
 
-### I am getting error `The feature is not available`
+### I am getting error `The feature is not available`, error code: `FeatureNotAvailable`
```
OrBaubergMicrosoft, line 151: what is the difference between templates/guides and docs/guides?
is the docs auto generated from the template? why?

DariuszPorowski: Yes, its auto generated from template. During generate, all willes are delete and generate new set of docs. This is how terraform docs tools works.

OrBaubergMicrosoft (approved):

PR #523, opened 2025-07-01
fix(workspace): add capacity state handling

### LINKED ISSUE #141: [BUG] Terraform wants to add new Fabric Enviroments that already exists if Capacity is Paused
### 🐛 What happened?

When the Fabric Capacity is paused, the terraform provider wants to add already existing items. Currently, I only have Fabric Environments deployed so this issue might also occur for other Fabric Items(Lakehouses, notebooks, etc.). 

If the Capacity is paused, the api will return 404 which might lead to the behavior of terraform wanting to deploy the item again
If this is the case, the terraform provider might need to check first if the Capacity is "running" before planning changes

### 🔬 How to reproduce?

1. deploy Fabric Environment using terraform provider
2. Pauze the Fabric Capacity
3. Run terraform plan/apply again

### 🏗️ Code Sample / Log

_No response_

### 📷 Screenshots

_No response_

### 📈 Expected behavior

data "fabric_capacity" "example" {
  display_name = var.capacity_name

  lifecycle {
    postcondition {
      condition = self.state == "Active"
      error_message = "Capacity is not Active"
    }
  }
}if data_source 

I added the following to create the expected behavior. This fixes the problem but it might be better to have the terraform provider to do this instead of a postcondition

0.1.0-beta.5

1.10

### 🌌 Environment (OS)

macOS

### 📎 Additional context

_No response_

### 🔰 Code of Conduct

DariuszPorowski:
# 📥 Pull Request

close #141

This pull request introduces enhancements to the `resourceWorkspace` implementation in `internal/services/workspace/resource_workspace.go`. The changes include adding support for capacity validation during workspace operations, improving error handling, and introducing a new helper method for capacity checks.

### Enhancements to workspace capacity handling:

* Added a `clientCapacity` field to `resourceWorkspace` for interacting with capacity-related operations. (`[internal/services/workspace/resource_workspace.goR35](diffhunk://#diff-87c99c17b488a2914da585da7b4f1eacb8b55b7439c0f8a2cff0c51753da2608R35)`)
* Enhanced the `get` method to validate workspace capacity by introducing a `capacityID` variable and delegating capacity checks to a new helper method, `getCapacity`. (`[[1]](diffhunk://#diff-87c99c17b488a2914da585da7b4f1eacb8b55b7439c0f8a2cff0c51753da2608R467)`, `[[2]](diffhunk://#diff-87c99c17b488a2914da585da7b4f1eacb8b55b7439c0f8a2cff0c51753da2608L487-R544)`)
* Implemented the `getCapacity` method to verify if the workspace's capacity is in an active state, ensuring stability and preventing potential damage from inactive capacities. It also handles cases where the capacity ID cannot be found, adding detailed error diagnostics. (`[internal/services/workspace/resource_workspace.goL487-R544](diffhunk://#diff-87c99c17b488a2914da585da7b4f1eacb8b55b7439c0f8a2cff0c51753da2608L487-R544)`)


internal/services/workspace/helpers.go
```diff
@@ -0,0 +1,77 @@
...
+			fmt.Sprintf(common.ErrorWorkspaceNotSupportedDetails, string(fabcore.WorkspaceTypeAdminWorkspace)),
+		)
+
+		return diags
+	default:
+		return nil
+	}
+}
+
+func getCapacity(ctx context.Context, client *fabcore.CapacitiesClient, capacityID *string) diag.Diagnostics {
```
badeamarjieh: seems like this is `validateCapacityState` or `ensureActiveCapacityState` instead of `getCapacity`

internal/services/workspace/helpers.go
```diff
@@ -0,0 +1,77 @@
...
+						"Fabric Capacity is NOT in Active state. Inactive Capacity may cause unrecoverable damage. Please ensure the Capacity is in Active state before continuing.",
+					)
+
+					return diags
+				}
+
+				return nil
+			}
+		}
+	}
```
badeamarjieh, line 69: why are we using `NewListCapacitiesPager` instead of `ListCapacities`? performance optimization?

internal/services/workspace/data_workspace_test.go
```diff
@@ -20,15 +20,17 @@ import (
...
-	entityTypePersonal := fakes.NewRandomWorkspaceInfoWithType(fabcore.WorkspaceTypePersonal)
-	entityTypeAdmin := fakes.NewRandomWorkspaceInfoWithType(fabcore.WorkspaceTypeAdminWorkspace)
+	capacity := fakes.NewRandomCapacity()
+	entity := fakes.NewRandomWorkspaceInfo(capacity.ID)
+	entityTypePersonal := fakes.NewRandomWorkspaceInfoWithType(fabcore.WorkspaceTypePersonal, nil)
+	entityTypeAdmin := fakes.NewRandomWorkspaceInfoWithType(fabcore.WorkspaceTypeAdminWorkspace, nil)
 
-	fakes.FakeServer.Upsert(fakes.NewRandomWorkspaceInfo())
+	fakes.FakeServer.Upsert(capacity)
+	fakes.FakeServer.Upsert(fakes.NewRandomWorkspaceInfo(capacity.ID))
```
badeamarjieh: do we want to have tests for both workspace with and without capacity?

internal/services/workspace/helpers.go
DariuszPorowski: yes, instead of getting all values, checking page by page

internal/services/workspace/data_workspace_test.go
DariuszPorowski: Makes sense, will add

internal/services/workspace/helpers.go
DariuszPorowski: aggree, will change

badeamarjieh (approved):

PR #560, opened 2025-07-28
fix(wellknown): update wellknown script to assign OneLake Data Access…

CatalinDabuleanu:
… Security and create LakehouseRS

# 📥 Pull Request

Add OneLake and LakehouseRS to wellknown variables


tools/scripts/Set-WellKnown.ps1
```diff
@@ -1135,6 +1135,46 @@ $wellKnown['KQLDatabase'] = @{
...
+
+    [Parameter(Mandatory = $true)]
+    [string]$LakehouseID
+  )
+
+  $results = Invoke-FabricRest -Method 'GET' -Endpoint "workspaces/$WorkspaceId/items/$LakehouseID/dataAccessRoles"
+
+  $result = $results.Response.value
+  if (!$result) {
+    Write-Log -Message "OneLake Data Access Security feature is not enabled for Lakehouse. Turn this feature on in the Fabric Admin Portal." -Level 'WARN'
```
OrBaubergMicrosoft: you showed me something needs to be done manually one the UI, not in the admin portal, what am I missing?

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1135,6 +1135,46 @@ $wellKnown['KQLDatabase'] = @{
...
+    return
+  }
+
+  Write-Log -Message "Assigning OneLake Data Access Security for Lakehouse: $($LakehouseID)"
+  $wellKnown['OneLakeDataAccessSecurity'] = @{
+    value = $result
```
OrBaubergMicrosoft: the result seems to be a large json object, do you need all of it?
why are you saving it?

OrBaubergMicrosoft: how are they created? you said that when you configure it manually, there is a default one, is there a default name you can filter by?

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1647,6 +1687,8 @@ $wellKnown['Subfolder'] = @{
   parentFolderId = $subFolder.parentFolderId
 }
 
+Set-OneLakeDataAccessSecurityItem -WorkspaceID $wellKnown['WorkspaceDS'].id -LakehouseID $wellKnown['Lakehouse'].id
```
OrBaubergMicrosoft: lakehouse RS  instead of lakehouse:)

CatalinDabuleanu: This is actually expected to be Lakehouse instead of LakehouseRS in this context, because we use onelake data access security in wellknown to check if the values retrieved from list are the ones we expect. And that is a datasource context.

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1344,6 +1362,7 @@ $results = Invoke-FabricRest -Method 'GET' -Endpoint "workspaces/$($wellKnown['W
 $result = $results.Response.data | Where-Object { $_.name -eq 'publicholidays' }
 if (!$result) {
   Write-Log -Message "!!! Please go to the Lakehouse and manually run 'Start with sample data' -> 'Public holidays' to populate the data !!!" -Level 'ERROR' -Stop $false
+  Write-Log -Message "OneLake Data Access Security feature is not enabled for Lakehouse. Please go to the Lakehouse and manually turn on this feature by clicking 'Manage OneLake data access'." -Level 'ERROR' -Stop $false
```
OrBaubergMicrosoft: please note here that the lakehouse is in workspace ds and the other one is in RS for less end user confusion

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1199,6 +1199,24 @@ $wellKnown['KQLDatabase'] = @{
...
+  $wellKnown['LakehouseRS'] = @{
+    id          = $item.id
+    displayName = $item.displayName
+    description = $item.description
+  }
+}
+
+#Set Lakehouse for WorkspaceRS
+Set-LakehouseResourceItem -DisplayName $displayName
```
OrBaubergMicrosoft: I think that this can be done outside of a set-... function
you but this is up to you
btw, what is displayName local var?

PR #573, opened 2025-08-05
feat(wellknown): add warehouse snapshot support

catalina-gav:
# 📥 Pull Request

Implement support for warehouse snapshot in WellKnown script


tools/scripts/Set-WellKnown.ps1
```diff
@@ -1741,6 +1745,20 @@ $wellKnown['Subfolder'] = @{
   parentFolderId = $subFolder.parentFolderId
 }
 
+# Create Warehouse Snapshot if not exists
+$displayNameTemp = "${displayName}_$($itemNaming['WarehouseSnapshot'])"
+$creationPayload = @{
+  parentWarehouseId = $wellKnown['Warehouse'].id
+}
+
+$warehouseSnapshot = Set-FabricItem -DisplayName $displayNameTemp -WorkspaceId $wellKnown['WorkspaceDS'].id -Type 'WarehouseSnapshot' -CreationPayload $creationPayload
```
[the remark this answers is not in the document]
catalina-gav: workspaceId is not an existing property in $wellKnown['Warehouse']

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1741,6 +1745,25 @@ $wellKnown['Subfolder'] = @{
   parentFolderId = $subFolder.parentFolderId
 }
 
+# Create Warehouse Snapshot if not exists
+if (-not $wellKnown.ContainsKey('Warehouse') -or -not $wellKnown['Warehouse'] -or -not $wellKnown['Warehouse'].id) {
+  Write-Log -Message "Warehouse not found or missing 'id'. Cannot create Warehouse Snapshot." -Level 'WARN'
```
OrBaubergMicrosoft, line 1750: if there is no warehouse, we should not create the snapshot

catalina-gav: We don't. The creation is in the else block. We just write that log if there is no warehouse

OrBaubergMicrosoft (approved):

PR #605, opened 2025-08-21
test: temp skip tests due to SPN auth issue

DariuszPorowski:
- test: temporarily skip tests due to SPN auth issue
- bump terraform to the latest: 1.13


BSamodien (approved):
LG

PR #630, opened 2025-08-30
feat(wellknown): add external data share to wellknown

CatalinDabuleanu:
# 📥 Pull Request

Implement External Data Share to Wellknown JSON

### DIFF
--- a/tools/scripts/Set-WellKnown.ps1
+++ b/tools/scripts/Set-WellKnown.ps1
@@ -82,7 +82,7 @@ function Set-ExternalDataShare {
   )
 
   $results = Invoke-FabricRest -Method 'GET' -Endpoint "workspaces/$WorkspaceId/items/$ItemId/externalDataShares"
-  $result = $results.Response.value | Where-Object { $_.recipient.userPrincipalName -eq $RecipientUserPrincipalName }
+  $result = $results.Response.value | Where-Object { $_.recipient.userPrincipalName -eq $RecipientUserPrincipalName } | Select-Object -First 1
   if (!$result) {
     Write-Log -Message "Creating External Data Share for Lakehouse: $ItemId" -Level 'WARN'
     $TABLES_PATH = "Tables"


tools/scripts/Set-WellKnown.ps1
```diff
@@ -69,6 +69,20 @@ function Import-ModuleIfNotImported {
...
+function Set-ExternalDataShare {
+  param (
+    [Parameter(Mandatory = $true)]
+    [string]$WorkspaceId,
+
+    [Parameter(Mandatory = $true)]
+    [string]$ItemId
+  )
+
+  $results = Invoke-FabricRest -Method 'GET' -Endpoint "workspaces/$WorkspaceId/items/$ItemId/externalDataShares"
```
OrBaubergMicrosoft, line 84: if there always a default EDS?

tools/scripts/Set-WellKnown.ps1
```diff
@@ -69,6 +69,32 @@ function Import-ModuleIfNotImported {
...
+  $result = $results.Response.value | Where-Object { $_.recipient.userPrincipalName -eq "test@example.com" }
+  if (!$result) {
+    Write-Log -Message "Creating External Data Share for Lakehouse: $ItemId" -Level 'WARN'
+    $payload = @{
+      paths     = @(
+        "Tables/publicholidays"
+      )
+      recipient = @{
+        userPrincipalName = "test@example.com"
```
catalina-gav: can't you use the same userPrincipalName we use everywhere in the script:
$userPrincipalName = $azContext.Account.Id, the id actually contains the test user's email

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1462,6 +1488,13 @@ $wellKnown['Datamart'] = @{
   description = if ($result) { $result.description } else { '' }
 }
 
+$externalDataShares = Set-ExternalDataShare -WorkspaceId $workspace.id -ItemId $wellKnown['Lakehouse'].id
```
catalina-gav: verify if the wellknown contains the Lakehouse key and that it is populated, see how it's done for warehouse for eg

catalina-gav (approved):

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1519,6 +1549,18 @@ $wellKnown['Datamart'] = @{
   description = if ($result) { $result.description } else { '' }
 }
 
+if (-not $wellKnown.ContainsKey('Lakehouse') -or -not $wellKnown['Lakehouse'].id) {
+  Write-Log -Message "Lakehouse not found or missing 'id'. Cannot create External Data Share." -Level 'WARN'
+}
+else {
+  $externalDataShares = Set-ExternalDataShare -WorkspaceId $workspace.id -ItemId $wellKnown['Lakehouse'].id -RecipientUserPrincipalName $azContext.Account.Id
+  $wellKnown['ExternalDataShare'] = @{
+    id          = $externalDataShares[0].id
```
catalina-gav: shouldnt the set return just the filtered result or created result? why is it being handled as an array?

tools/scripts/Set-WellKnown.ps1
```diff
@@ -69,6 +69,36 @@ function Import-ModuleIfNotImported {
...
+
+    [Parameter(Mandatory = $true)]
+    [string]$ItemId,
+
+    [Parameter(Mandatory = $true)]
+    [string]$RecipientUserPrincipalName
+  )
+
+  $results = Invoke-FabricRest -Method 'GET' -Endpoint "workspaces/$WorkspaceId/items/$ItemId/externalDataShares"
+  $result = $results.Response.value | Where-Object { $_.recipient.userPrincipalName -eq $RecipientUserPrincipalName }
```
catalina-gav: `$result = $results.Response.value | Where-Object { $_.recipient.userPrincipalName -eq $RecipientUserPrincipalName } | Select-Object -First 1`
To make sure you always get 1 result since you re only filtering based on userPrincipalName

CatalinDabuleanu: We use this $results.Response.value | Where-Object { $_. [...] -eq } 30 times in our code, we dont use -First 1 in any of those places.

catalina-gav: We do, especially when filtering by non-unique parameters:

Image

Image

When filtering by displayName which is unique, there is no point in using it yes

PR #634, opened 2025-09-02
fix(shortcut): resource acc test should use WorkspaceRS

catalina-gav:
# 📥 Pull Request

Use WorkspaceRS for shortcut resource creation acc test instead of workspace DS

### DIFF
--- a/internal/services/shortcut/base_test.go
+++ b/internal/services/shortcut/base_test.go
@@ -4,7 +4,28 @@
 package shortcut_test
 
 import (
+	"testing"
+
+	at "github.com/dcarbone/terraform-plugin-framework-utils/v3/acctest"
+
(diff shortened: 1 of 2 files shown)


internal/services/shortcut/base_test.go
```diff
@@ -4,7 +4,28 @@
...
+	"testing"
+
+	at "github.com/dcarbone/terraform-plugin-framework-utils/v3/acctest"
+
 	"github.com/microsoft/terraform-provider-fabric/internal/services/shortcut"
+	"github.com/microsoft/terraform-provider-fabric/internal/testhelp"
 )
 
 var itemTypeInfo = shortcut.ItemTypeInfo
+
```
OrBaubergMicrosoft, line 16: why is this needed here?

Issue #620, opened 2025-08-27
[BUG] Shortcuts produce inconsistent results for Schema shortcuts


cseHdz:
### 🐛 What happened?

When creating a onelake shortcut when the path variable ends in forward slash ("Tables/schema_name/") and the name variable is an empty string (""), the provider successfully creates a Schema shortcut.

However, this produces an inconsistent result after apply as the registry expects the name to be an empty string("") and the provider correctly returns the name of the schema.

Alternatively, if passing schema_name in the name variable, then it creates a schema shortcut under the Lakehouse schema which creates `schema_name / Unidentified / schema_name / tables`.

### 🔬 How to reproduce?

Create a resource with the table_name variable set to an empty string ("")

``` 
# main.tf
resource "fabric_shortcut" "onelake_lh" { 
  workspace_id = var.workspace_id
  item_id      = var.lakehouse_id
  name         = var.table_name
  path         = "Tables/${var.schema_name}"

  target = {
    onelake = {
      workspace_id = var.source_workspace_id
      item_id      = var.source_lakehouse_id
      path         = "Tables/${var.schema_name}/${var.table_name}"
    }
  }
}
```

### 🏗️ Code Sample / Log

  │ Error: Provider produced inconsistent result after apply
  │ 
  │ When applying changes to
  │ fabric_shortcut.onelake_lh["ws-prd-edp-bronze-schema-"], provider
  │ "provider[\"registry.terraform.io/microsoft/fabric\"]" produced an
  │ unexpected new value: .name: was cty.StringVal(""), but now
  │ cty.StringVal("schema").
  │ 
  │ This is a bug in the provider, which should be reported in the provider's
  │ own issue tracker.

### 📈 Expected behavior

When the name variable is set to an empty string (""), the expected name should be the name of the schema as opposed to an empty string (""). This will enable creating schema shortcuts.

To be safe, add validation on the target variable to ensure the "path" parameter also ends in a forward slash (/) if the name variable is an empty string (""). 

### 🌌 Environment (Provider Version)

1.3.0 (preview)

### 🌌 Environment (Terraform Version)

1.13.0

### 🌌 Environment (OS)

macOS

### 🔰 Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

cseHdz:
The correct behaviour for schema shortcuts can be achieved through the existing providers with the following variables:

name = schema_name
path = "Tables" OR path ="Tables/"

[...]

PR #640, opened 2025-09-04
fix(shortcut): name validation to avoid empty name

catalina-gav:
# 📥 Pull Request
Solve #620 
Add shortcut name validation - name is required and it must contain at least one non whitespace character


internal/services/shortcut/schema.go
```diff
@@ -109,6 +109,17 @@ func itemSchema(isList bool) superschema.Schema { //revive:disable-line:flag-par
 					PlanModifiers: []planmodifier.String{
 						stringplanmodifier.RequiresReplace(),
 					},
+					Validators: []validator.String{
+						stringvalidator.LengthAtLeast(1),
+						stringvalidator.RegexMatches(
+							regexp.MustCompile(`\S`),
+							"Name must contain at least one non-whitespace character.",
+						),
+						stringvalidator.RegexMatches(
```
OrBaubergMicrosoft: IMO, this is too specific, let it fail on the API side

PR #647, opened 2025-09-09
feat(workspace_git): add ConfiguredConnection credential support and update SP docs

badeamarjieh:
# 📥 Pull Request

- add ConfiguredConnection support for "fabric_workspace_git" AzDO provider, and update SP support documentation only when git credentials is ConfiguredConnection.

### DIFF
--- a/.changes/unreleased/changed-20250910-072503.yaml
+++ b/.changes/unreleased/changed-20250910-072503.yaml
@@ -1,0 +1,5 @@
+kind: changed
+body: Enabled SPN support for `fabric_workspace_git` only for `ConfiguredConnection` credentials.
+time: 2025-09-10T07:25:03.081601685Z
+custom:
+    Issue: "647"
--- a/docs/data-sources/workspace_git.md
+++ b/docs/data-sources/workspace_git.md
@@ -4,7 +4,7 @@ page_title: "fabric_workspace_git Data Source - terraform-provider-fabric"
 subcategory: ""
 description: |-
   The Workspace Git data-source allows you to retrieve details about a Fabric Workspace Git https://learn.microsoft.com/fabric/cicd/git-integration/intro-to-git-integration.
-  -> This data-source supports Service Principal authentication only when git_provider_details.git_provider_type is "GitHub".
+  -> This data-source supports Service Principal authentication only when git_credentials.source is "ConfiguredConnection".
   ~> This data-source is in preview. To access it, you must explicitly enable the preview mode in the provider level configuration.
 ---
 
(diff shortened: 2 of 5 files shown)


lorneeaston:
Yes!

internal/services/workspacegit/models.go
```diff
@@ -132,8 +132,36 @@ func (to *requestGitConnect) set(ctx context.Context, from resourceWorkspaceGitM
...
+		switch gitGitCredentialsSource {
+		case fabcore.GitCredentialsSourceAutomatic:
+			reqGitCredentials = &fabcore.AutomaticGitCredentials{
+				Source: &gitGitCredentialsSource,
+			}
+
+		case fabcore.GitCredentialsSourceConfiguredConnection:
+			reqGitCredentials = &fabcore.ConfiguredConnectionGitCredentials{
+				Source:       &gitGitCredentialsSource,
+				ConnectionID: gitGitCredentials.ConnectionID.ValueStringPointer(),
```
OrBaubergMicrosoft, line 155: comment to self: check in schema that if source is GitCredentialsSourceConfiguredConnection then ConnectionID is mandatory, what happens if not?

OrBaubergMicrosoft: it is, resolving comment

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1236,16 +1237,24 @@ $wellKnown['KQLDatabase'] = @{
...
   id          = $item.id
   displayName = $item.displayName
   description = $item.description
 }
 
+$results = Invoke-FabricRest -Method 'GET' -Endpoint "workspaces/$($wellKnown['WorkspaceRS'].id)/lakehouses/$($wellKnown['LakehouseRS']['id'])/tables"
+$result = $results.Response.data | Where-Object { $_.name -eq 'publicholidays' }
+if (!$result) {
+  Write-Log -Message "!!! Please go to the Lakehouse inside Workspace: $($wellKnown['WorkspaceRS'].displayName) and manually run 'Start with sample data' -> 'Public holidays' to populate the data !!!" -Level 'ERROR' -Stop $false
+  Write-Log -Message "OneLake Data Access Security feature is not enabled for Lakehouse. Please go to the Lakehouse inside Workspace: $($wellKnown['WorkspaceRS'].displayName) and manually turn on this feature by clicking 'Manage OneLake data access'." -Level 'ERROR' -Stop $false
```
OrBaubergMicrosoft, line 1253: you are coupling here lakehouse table being populated with data access security being enabled
meaning, in case Access Security is disabled but the table is populated, this will not notify us that we should do something

OrBaubergMicrosoft: fix on another pr

PR #676, opened 2025-10-01
feat(domain): use non preview domains API

catalina-gav:
# 📥 Pull Request

Use the on preview models and methods from the sdk for domain data source and resource


internal/services/domain/resource_domain.go
```diff
@@ -96,28 +96,12 @@ func (r *resourceDomain) Create(ctx context.Context, req resource.CreateRequest,
...
-		reqUpdate.set(plan)
-
-		respUpdate, err := r.client.UpdateDomainPreview(ctx, state.ID.ValueString(), true, reqUpdate.UpdateDomainRequestPreview, nil)
-		if resp.Diagnostics.Append(utils.GetDiagsFromError(ctx, err, utils.OperationCreate, nil)...); resp.Diagnostics.HasError() {
-			return
-		}
-
-		state.set(respUpdate.DomainPreview)
-	}
+	state.set(respCreate.Domain)
```
badeamarjieh, line 105: to set the `defaultLabelId`, you need to call the Update Domain API - it is not settable in the Create Domain API

[Create Request Body](https://learn.microsoft.com/en-us/rest/api/fabric/admin/domains/create-domain?tabs=HTTP#request-body)
[Update Request Body](https://learn.microsoft.com/en-us/rest/api/fabric/admin/domains/update-domain?tabs=HTTP#request-body)

internal/services/domain/resource_domain.go
```diff
@@ -196,12 +180,12 @@ func (r *resourceDomain) Update(ctx context.Context, req resource.UpdateRequest,
 
 	reqUpdate.set(plan)
```
badeamarjieh, line 210: also, upon Update, if we are removing the `defaultLabelId`, preprocessing should be done before we call the Update API so set it to null, as the API requests an empty GUID: https://learn.microsoft.com/en-us/rest/api/fabric/admin/domains/update-domain?tabs=HTTP#request-body

internal/services/domain/resource_domain.go
```diff
@@ -103,6 +104,19 @@ func (r *resourceDomain) Create(ctx context.Context, req resource.CreateRequest,
 
 	state.set(respCreate.Domain)
```
badeamarjieh, line 105:
```
   state.set(respCreate.Domain)

    // Save state immediately after resource creation
    // This ensures Terraform can track the resource even if subsequent operations fail
    resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
    if resp.Diagnostics.HasError() {
        return
    }

    // The Create API doesn't support default_label_id, we need to call Update to set it
    if !plan.DefaultLabelID.IsNull() && !plan.DefaultLabelID.IsUnknown() {
        tflog.Debug(ctx, "CREATE", map[string]any{
            "action":           "setting default_label_id via Update API",
            "default_label_id": plan.DefaultLabelID.ValueString(),
        })

        var reqUpdate requestUpdateDomain
        reqUpdate.set(plan)

        respUpdate, err := r.client.UpdateDomain(ctx, state.ID.ValueString(), ItemTypeInfo.IsPreview, reqUpdate.UpdateDomainRequest, nil)
        if resp.Diagnostics.Append(utils.GetDiagsFromError(ctx, err, utils.OperationUpdate, nil)...); resp.Diagnostics.HasError() {
            // State already saved — resource exists, Terraform knows about it
            // Next apply will show drift and retry the Update
            return
        }

        state.set(respUpdate.Domain)
        resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
    }

    tflog.Debug(ctx, "CREATE", map[string]any{"action": "end"})
```

internal/services/domain/resource_domain.go
badeamarjieh, line 223: why is this needed? the plan is set in the beginning

badeamarjieh: redundant

.changes/unreleased/changed-20260122-073612.yaml
```diff
@@ -1,5 +1,5 @@
 kind: changed
-body: add support for `fabric_domain` GA APIs, remove preview
+body: add support for `fabric_domain` GA APIs, set preview to false
```
badeamarjieh: Removed preview for `fabric_domain`

badeamarjieh: its not about setting preview to false - users can't choose between preview and non-preview domain. but now they no longer need the preview = true config to use the domains (but maybe they still need it for other preview resources/data-sources that do require it)

PR #682, opened 2025-10-06
chore: codebase maintenance

DariuszPorowski:
- codebase lint to meet a new golangci-lint rules
- re-enable GH test
- extend timeout to 40 minutes


BSamodien (approved):
LGTM

Issue #679, opened 2025-10-01
[RS] Add Tags Resource


CatalinDabuleanu:
### 📝 Description

When I am automating and managing data analytics infrastructure using Microsoft Fabric across multiple environments, I want to get details of Tags as code in Terraform.
To consistently and securely manage Tag as code in Terraform while automating and managing data analytics infrastructure using Microsoft Fabric across multiple environments.

### 🔬 Details / References

- Resource Name: `fabric_tag`
- API documentation: 
  - Bulk Create Tags: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/bulk-create-tags?tabs=HTTP
  - Update Tag: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/update-tag?tabs=HTTP
  - Delete Tag: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/delete-tag?tabs=HTTP
- Estimated complexity/effort: easy
- Related resources/data-sources: 
  - List Tags: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/list-tags?tabs=HTTP

### 🚧 Potential Terraform Configuration / Desired Solution

```terraform
resource "fabric_tag" "example" {
 tags = [
 {
   display_name = "example1"
 },
 {
   display_name = "example2"
 }
]
}
```

### ✅ Definition of Done

- [ ] Data Transfer Objects (DTOs)
- [ ] Resource Implementation
- [ ] Resource Added to Provider
- [ ] Unit Tests for Happy path
- [ ] Unit Tests for Error path
- [ ] Acceptance Tests
- [ ] Example in the ./examples folder
- [ ] Schema documentation in code
- [ ] Updated auto-generated provider docs with `task docs`

### 🔰 Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

PR #688, opened 2025-10-12
feat(tags): Onboard new fabric_tag data source and resource

### LINKED ISSUE #678: [DS] Add Tags Data Source
### 📝 Description

When I am automating and managing data analytics infrastructure using Microsoft Fabric across multiple environments, I want to get details of Tags as code in Terraform.
To consistently and securely manage Tags as code in Terraform while automating and managing data analytics infrastructure using Microsoft Fabric across multiple environments.

### 🔬 Details / References

- Data Source Name: `fabric_tags`
- API documentation: 
  - List Tags: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/list-tags?tabs=HTTP
- Estimated complexity/effort: easy
- Related resources/data-sources: 
  - Bulk Create Tags: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/bulk-create-tags?tabs=HTTP
  - Delete Tag: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/delete-tag?tabs=HTTP
  - Update Tag: https://learn.microsoft.com/en-us/rest/api/fabric/admin/tags/update-tag?tabs=HTTP

### 🚧 Potential Terraform Configuration / Desired Solution

```terraform
data "fabric_tags" "example" {
}
```

```terraform
data "fabric_tags" "example" {
    display_name = "test"
}
```

```terraform
data "fabric_tags" "example" {
     id = guid
}
```

### 📎 Additional context

_No response_

### ☑️ Acceptance Criteria

_No response_

### ✅ Definition of Done

- [ ] Data Transfer Objects (DTOs)
- [ ] Data-Source Implementation
- [ ] Data-Source Added to Provider
- [ ] Unit Tests for Happy path
- [ ] Unit Tests for Error path
- [ ] Acceptance Tests
- [ ] Example in the ./examples folder
- [ ] Schema documentation in code
- [ ] Updated auto-generated provider docs with `task docs`

### 🔰 Code of Conduct

CatalinDabuleanu:
# 📥 Pull Request

Onboard new `fabric_tags` data source and resource.

This closes #678 and #679


internal/services/tags/data_tag.go
```diff
@@ -0,0 +1,185 @@
...
+		for _, entity := range page.Value {
+			if *entity.ID == model.ID.ValueString() {
+				return model.set(ctx, entity)
+			}
+		}
+	}
+
+	diags.AddError(
+		common.ErrorReadHeader,
+		"Unable to find Tag with 'display_name': "+model.DisplayName.ValueString(),
```
catalina-gav: with 'id', model.ID.ValueString()

internal/services/tags/data_tag.go
```diff
@@ -0,0 +1,185 @@
...
+		page, err := pager.NextPage(ctx)
+		if diags := utils.GetDiagsFromError(ctx, err, utils.OperationList, nil); diags.HasError() {
+			return diags
+		}
+
+		for _, entity := range page.Value {
+			if *entity.DisplayName == model.DisplayName.ValueString() {
+				model.ID = customtypes.NewUUIDPointerValue(entity.ID)
+
+				return d.getByID(ctx, model)
```
catalina-gav: Consider using return model.set(ctx, entity) directly here instead of getByID(ctx, model). Since there's no dedicated "get tag by ID" API endpoint, getByID just iterates through the same list again, which is redundant when you already have the entity.

internal/services/tags/fake_test.go
```diff
@@ -0,0 +1,125 @@
...
+		errItemNotFound := fabcore.ErrItem.ItemNotFound.Error()
+
+		if _, ok := fakeTagStore[tagID]; !ok {
+			errResp.SetError(fabfake.SetResponseError(http.StatusNotFound, errItemNotFound, "Item not found"))
+			resp.SetResponse(http.StatusNotFound, fabadmin.TagsClientUpdateTagResponse{}, nil)
+
+			return resp, errResp
+		}
+
+		tag := fabadmin.TagInfo{
```
catalina-gav, line 57: I think you could use the request parameter (UpdateTagRequest) directly to form the updated TagInfo, instead of the outer tag variable. That way the fake properly simulates the real API behavior. Maybe rename the variable to updateTagRequest

internal/services/tags/fake_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+		fakeTestUpsert(tag)
+		resp.SetResponse(http.StatusOK, fabadmin.TagsClientUpdateTagResponse{Tag: returnTag}, nil)
+
+		return resp, errResp
+	}
+}
+
+func fakeBulkCreateTagsFunc(
+	body fabadmin.CreateTagsRequest,
```
catalina-gav: Same issue - remove outer body param and use the inner request parameter directly. maybe rename it to createTagsRequest

internal/services/tags/data_tag_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/microsoft/terraform-provider-fabric/internal/common"
+	"github.com/microsoft/terraform-provider-fabric/internal/framework/customtypes"
+	"github.com/microsoft/terraform-provider-fabric/internal/testhelp"
+	"github.com/microsoft/terraform-provider-fabric/internal/testhelp/fakes"
+)
+
+var testDataSourceItemFQN, testDataSourceItemHeader = testhelp.TFDataSource(common.ProviderTypeName, itemTypeInfo.Type, "test")
+
+func TestUnit_TagDataSource(t *testing.T) {
+	randomEntity := NewRandomTag()
```
catalina-gav: we usually use just entity for randomEntities in tests, so maybe could rename to simply entity

internal/services/tags/models.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/*
+RESOURCE
+*/
+
+type resourceTagsModel struct {
+	baseTagModel
+
+	Tags supertypes.ListNestedObjectValueOf[baseTagModel] `tfsdk:"tags"`
```
catalina-gav: how will the update work when we create the resource having multiple tags? how will it work considering that for update we always use the baseTagModel. shouldn't one resource instance correspond to a single managed object? maybe even though we have  only a bulk create API, the fabric_tag resource should manage a single tag?

internal/services/tags/resource_tag_test.go
```diff
@@ -0,0 +1,140 @@
...
+				},
+			),
+			Check: resource.ComposeAggregateTestCheckFunc(
+				resource.TestCheckResourceAttr(testResourceItemFQN, "tags.0.display_name", entity1DisplayName),
+				resource.TestCheckResourceAttr(testResourceItemFQN, "tags.1.display_name", entity2DisplayName),
+				resource.TestCheckResourceAttrSet(testResourceItemFQN, "tags.0.id"),
+				resource.TestCheckResourceAttrSet(testResourceItemFQN, "tags.1.id"),
+			),
+		},
+		// Update and Read
```
catalina-gav: following the question on the models, how would this test the update method and not the create method?

internal/services/tags/schema_data_tag.go
```diff
@@ -0,0 +1,109 @@
...
+		DataSource: superschema.SchemaDetails{
+			MarkdownDescription: fabricitem.NewDataSourceMarkdownDescription(ItemTypeInfo, isList),
+		},
+		Attributes: map[string]superschema.Attribute{
+			"id": superschema.SuperStringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "The " + ItemTypeInfo.Name + " ID.",
+					CustomType:          customtypes.UUIDType{},
+				},
+				Resource: &schemaR.StringAttribute{
```
catalina-gav, line 49: if we have the sepparate schema_resource_tag, why do we handle resources here as well and not only data sources?

internal/services/tags/data_tag.go
CatalinDabuleanu: image
image

This is standard across our artifacts. DisplayName is unique, so its enough to consider it as an unique identifier

internal/services/tags/fake_test.go
CatalinDabuleanu: This Update method works just as a placeholder currently until a documentation bug is solved. issue has already been raised to the correct team, they are investigating. will update this piece of code once the documentation is properly updated

internal/services/tags/models.go
CatalinDabuleanu: Not sure I understand the question, the 2 examples in examples folder and acceptance tests explain the scenarios.

If you provide a single object, an update will be triggered if the item is found, if multiple objects are provided, a create is going to be triggered. 

Create doesnt care about the single object and Update doesnt care about the array of tags.

internal/services/tags/resource_tag_test.go
CatalinDabuleanu: As said above, setters in Create and Update do the trick and make the difference between operations

internal/services/tags/data_tag.go
catalina-gav: I meant that the log should correspond to the method it is being called from, in this case getByID, so should log the id, not display name. In most places in the provider we do not display this error for getById because it's an API call and just return the diags error message if it's the case.

internal/services/tags/data_tag.go
```diff
@@ -0,0 +1,185 @@
...
+	tflog.Debug(ctx, "READ", map[string]any{
+		"action": "end",
+	})
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+func (d *dataSourceTag) getByID(ctx context.Context, model *dataSourceTagModel) diag.Diagnostics {
```
catalina-gav, line 126: since there is no Get Tag API you could have a single get method where you handle both cases of get: by id or display_name, as it is done in data_capacity for example

internal/services/tags/models.go
catalina-gav: @OrBaubergMicrosoft

.changes/unreleased/added-20251012-225400.yaml
```diff
@@ -0,0 +1,5 @@
+kind: added
+body: Onboard new `fabric_tag` Data Source
```
OrBaubergMicrosoft, line 2: Q regarding the name
we are creating multiple tags in the same resource, maybe we should rename to tags

examples/resources/fabric_tag/resource.tf
```diff
@@ -0,0 +1,20 @@
...
+}
+
+#Update
+resource "fabric_tag" "example_update" {
+  id           = "00000000-0000-0000-0000-000000000000"
+  display_name = "example_updated"
+  scope = {
+    type = "Tenant"
+  }
+}
```
OrBaubergMicrosoft, line 20: this is misleading
you can create with scope, but update scope recreates the tag 
note that when updating a tag that you created, you will not specify the id

CatalinDabuleanu: Tenant is the default scope created if not provided explicitly in the Create.

.changes/unreleased/added-20251012-225400.yaml
CatalinDabuleanu: yes, but for update we are only doing it for 1 tag at a time. What do you think? should it still be called 'tags' or 'tag'?

examples/resources/fabric_tag/resource.tf
OrBaubergMicrosoft: This is still a weird way to update, I think that you can update just the display name in your example and not note the scope
keep a creation example with explicit scope :)

.changes/unreleased/added-20251012-225400.yaml
OrBaubergMicrosoft: This should be further discussed, will try to open think about it

OrBaubergMicrosoft: Q:
bulk create tags (create tag A)
bulk create tags (create tag B)
list will return A and B?
if that is the case, I think we should treat the resource as a single tag
Another Q
are display names of tags unique?
can you triage the APIs and let me know here?

internal/services/tags/data_tag.go
```diff
@@ -0,0 +1,177 @@
...
+	return d.get(ctx, true, model)
+}
+
+func (d *dataSourceTag) getByDisplayName(ctx context.Context, model *dataSourceTagModel) diag.Diagnostics {
+	return d.get(ctx, false, model)
+}
+
+func (d *dataSourceTag) get(ctx context.Context, byID bool, model *dataSourceTagModel) diag.Diagnostics {
```
OrBaubergMicrosoft, line 134: is this a common pattern?
byID bool switch byID { seems to me like a new one, what most of DS do?

internal/services/tags/data_tags_test.go
```diff
@@ -0,0 +1,53 @@
...
+func TestAcc_TagsDataSource(t *testing.T) {
+	resource.ParallelTest(t, testhelp.NewTestAccCase(t, nil, nil, []resource.TestStep{
+		// read
+		{
+			Config: at.CompileConfig(
+				testDataSourceItemsHeader,
+				map[string]any{},
+			),
+			Check: resource.ComposeAggregateTestCheckFunc(
+				resource.TestCheckResourceAttrSet(testDataSourceItemsFQN, "values.0.display_name"),
```
OrBaubergMicrosoft, line 49: can you check id and type here and above?

internal/services/tags/fake_test.go
OrBaubergMicrosoft: is this due to the tag vs admintag?
after the docs change, will we need to update it here as well?

internal/services/tags/models.go
OrBaubergMicrosoft: discussed it above in
https://github.com/microsoft/terraform-provider-fabric/pull/688#discussion_r2763915119
I am leaning towards changing the design from
resource "fabric_tag" "example" {
 tags = [
 {
   display_name = "example1"
 },
 {
   display_name = "example2"
 }
]
}
to 
resource "fabric_tag" "example" {
   display_name = "example1"
}

internal/services/tags/resource_tag.go
```diff
@@ -0,0 +1,279 @@
...
+	timeout, diags := state.Timeouts.Delete(ctx, r.pConfigData.Timeout)
+	if resp.Diagnostics.Append(diags...); resp.Diagnostics.HasError() {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	if state.ID.IsNull() {
+		tags, diags := state.Tags.Get(ctx)
```
OrBaubergMicrosoft: from my understanding, if we change the semantics of this resouce to be singular, this resouce will be less fragile

internal/services/tags/data_tag.go
CatalinDabuleanu: Its similar with what's already in code for Connection https://github.com/microsoft/terraform-provider-fabric/blob/06893153b9c24d01d6c539b4838ef4395d89a981/internal/services/connection/models.go#L119

internal/services/tags/fake_test.go
CatalinDabuleanu: Yes

.changes/unreleased/added-20251012-225400.yaml
CatalinDabuleanu: Yes, tag names should be unique

docs/resources/tag.md
```diff
@@ -0,0 +1,68 @@
...
+resource "fabric_tag" "example_scope" {
+  display_name = "example_scope"
+  scope = {
+    type = "Tenant"
+  }
+}
+
+#Update
+resource "fabric_tag" "example_update" {
+  id           = "00000000-0000-0000-0000-000000000000"
```
OrBaubergMicrosoft: id is not needed here

internal/services/tags/models.go
```diff
@@ -0,0 +1,220 @@
...
+		DisplayName: from.DisplayName.ValueStringPointer(),
+	})
+
+	if from.Scope.IsKnown() {
+		scope, diags := from.Scope.Get(ctx)
+		if diags.HasError() {
+			return diags
+		}
+
+		if scope.Type.ValueString() == string(fabadmin.TagScopeTypeTenant) {
```
OrBaubergMicrosoft, line 155: we switch on the scope  type multiple times, can it be refactored?

internal/services/tags/schema.go
```diff
@@ -0,0 +1,123 @@
...
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "The " + ItemTypeInfo.Name + " ID.",
+					CustomType:          customtypes.UUIDType{},
+					Computed:            true,
+				},
+				DataSource: &schemaD.StringAttribute{
+					Optional: !isList,
+				},
+				Resource: &schemaR.StringAttribute{
+					Optional: true,
```
OrBaubergMicrosoft: should it be computed?

internal/services/tags/schema.go
```diff
@@ -0,0 +1,123 @@
...
+							},
+						},
+					},
+					"type": superschema.StringAttribute{
+						Common: &schemaR.StringAttribute{
+							MarkdownDescription: "Scope Type.",
+							Computed:            true,
+						},
+						Resource: &schemaR.StringAttribute{
+							Optional: true,
```
OrBaubergMicrosoft: if a tag has scope, then shouldn't the type be required?

PR #699, opened 2025-10-21
feat(wellknown): add support for Tags in Set-Wellknown script

CatalinDabuleanu:
# 📥 Pull Request

Add support for Tags in Set-Wellknown script

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
           passphrase: ${{ secrets.GPG_PRIVATE_KEY_PASSPHRASE }}
 
       - name: 🚧 Setup Syft
-        uses: anchore/sbom-action/download-syft@deef08a0db64bfad603422135db61477b16cef56 # v0.22.1
+        uses: anchore/sbom-action/download-syft@28d71544de8eaf1b958d335707167c5f783590ad # v0.22.2
 
       - name: 🚀 Run GoReleaser
         uses: goreleaser/goreleaser-action@e435ccd777264be153ace6237001ef4d979d3a7a # v6.4.0


tools/scripts/Set-WellKnown.ps1
```diff
@@ -1715,6 +1715,30 @@ $definition = @{
   )
 }
 
+function Set-Tags {
+  $results = Invoke-FabricRest -Method 'GET' -Endpoint "admin/tags"
+  $result = $results.Response.value[0]
```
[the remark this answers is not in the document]
CatalinDabuleanu: It's fine to pick the first tag in the list

tools/scripts/Set-WellKnown.ps1
```diff
@@ -1800,6 +1800,30 @@ $definition = @{
...
+  if (-not $result) {
+    $payload = @{
```
OrBaubergMicrosoft, line 1807: The payload should be an array of tags but the payload does not seem to be an array, am I missing anything?

OrBaubergMicrosoft: I think it should be
$payload = @(
  @{ displayName = "Test Tag" }
)
because there is no createTagsRequest layer

POST https://api.fabric.microsoft.com/v1/admin/tags/bulkCreateTags

[
  {
    "displayName": "Finance 2024"
  },
  {
    "displayName": "HR 2024"
  }
]

OrBaubergMicrosoft: actually, seems to be a bug in the docs, approving the pr

OrBaubergMicrosoft (approved):

PR #754, opened 2025-12-03
feat(tenant_settings): onboard new datasource and resource …

CatalinDabuleanu:
# 📥 Pull Request

Onboard tenant settings for datasource and resource

### DIFF
--- a/internal/services/tenantsetting/resource_tenant_setting_test.go
+++ b/internal/services/tenantsetting/resource_tenant_setting_test.go
@@ -126,7 +126,6 @@ func TestAcc_TenantSettingsResource_CRUD(t *testing.T) {
 					"enabled_security_groups": []map[string]any{
 						{
 							"graph_id": securityGroupID,
-							"name":     securityGroupName,
 						},
 					},
 					"delete_behaviour": string(tenantsetting.NoChange),
@@ -153,7 +152,6 @@ func TestAcc_TenantSettingsResource_CRUD(t *testing.T) {
 					"enabled_security_groups": []map[string]any{
 						{
 							"graph_id": securityGroupID,
-							"name":     securityGroupName,
 						},
 					},
 					"delete_behaviour": string(tenantsetting.Disable),
--- a/internal/services/tenantsetting/schema.go
+++ b/internal/services/tenantsetting/schema.go
@@ -139,9 +139,6 @@ func itemSchema(isList bool) superschema.Schema { //revive:disable-line:flag-par
 							MarkdownDescription: "The name of the security group.",
 							Computed:            true,
 						},
-						Resource: &schemaR.StringAttribute{
-							Optional: true,
-						},
 					},
 				},
 			},
@@ -170,9 +167,6 @@ func itemSchema(isList bool) superschema.Schema { //revive:disable-line:flag-par
 						Common: &schemaR.StringAttribute{
 							MarkdownDescription: "The name of the security group.",
 							Computed:            true,
-						},
-						Resource: &schemaR.StringAttribute{
-							Optional: true,
 						},
 					},
 				},
--- a/docs/resources/tenant_setting.md
+++ b/docs/resources/tenant_setting.md
@@ -46,7 +46,7 @@ Required:
 
 - `graph_id` (String) The graph ID of the security group.
 
-Optional:
+Read-Only:
 
 - `name` (String) The name of the security group.
 
@@ -58,7 +58,7 @@ Required:
 
 - `graph_id` (String) The graph ID of the security group.
 
-Optional:
+Read-Only:
 
 - `name` (String) The name of the security group.
 


internal/services/tenantsetting/data_tenant_settings_test.go
```diff
@@ -0,0 +1,54 @@
...
+func TestAcc_TenantSettingsDataSource(t *testing.T) {
+	resource.ParallelTest(t, testhelp.NewTestAccCase(t, nil, nil, []resource.TestStep{
+		// read
+		{
+			Config: at.CompileConfig(
+				testDataSourceItemsHeader,
+				map[string]any{},
+			),
+			Check: resource.ComposeAggregateTestCheckFunc(
+				resource.TestCheckResourceAttrSet(testDataSourceItemsFQN, "values.0.setting_name"),
```
catalina-gav, line 54: maybe add checks for the other attributes and properties too

internal/services/tenantsettings/resource_tenant_settings.go
```diff
@@ -0,0 +1,265 @@
...
+	var state resourceTenantSettingsModel
+
+	if resp.Diagnostics.Append(req.State.Get(ctx, &state)...); resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !state.DeleteBehaviour.IsNull() && !state.DeleteBehaviour.IsUnknown() && state.DeleteBehaviour.ValueString() == string(Disable) {
+		var reqUpdate requestUpdateTenantSettings
+		disabled := false
+		reqUpdate.UpdateTenantSettingRequest.Enabled = &disabled
```
catalina-gav: could replace with one liner 
reqUpdate.UpdateTenantSettingRequest.Enabled = to.Ptr(false)

internal/services/tenantsetting/resource_tenant_setting_test.go
```diff
@@ -0,0 +1,148 @@
...
+	entity := testhelp.WellKnown()["TenantSettings"].(map[string]any)
+	settingName := entity["settingName"].(string)
+	enabled := entity["enabled"].(bool)
+
+	securityGroupEntity := testhelp.WellKnown()["SecurityGroup"].(map[string]any)
+	securityGroupID := securityGroupEntity["id"].(string)
+	securityGroupName := securityGroupEntity["name"].(string)
+
+	resource.Test(t, testhelp.NewTestAccCase(t, &testResourceItemFQN, nil, []resource.TestStep{
+		// Update
```
catalina-gav, line 144: doesn't this only test the Create func from resource_tenant_settings? maybe also add a test case to update the resource again so it also tests the Update func

internal/services/tenantsettings/models.go
```diff
@@ -0,0 +1,242 @@
...
+				GraphID: sg.GraphID.ValueStringPointer(),
+				Name:    sg.Name.ValueStringPointer(),
+			}
+			slice = append(slice, securityGroup)
+		}
+		to.EnabledSecurityGroups = slice
+	}
+
+	if !from.ExcludedSecurityGroups.IsNull() && !from.ExcludedSecurityGroups.IsUnknown() {
+		sgs, diags := from.EnabledSecurityGroups.Get(ctx)
```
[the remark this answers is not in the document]
catalina-gav:
```suggestion
		sgs, diags := from.ExcludedSecurityGroups.Get(ctx)
```

internal/services/tenantsettings/models.go
```diff
@@ -0,0 +1,289 @@
...
+	for _, entity := range from {
+		if entity.SettingName != nil && *entity.SettingName == to.SettingName.ValueString() {
+			return to.set(ctx, entity)
+		}
+	}
+	var diags diag.Diagnostics
+
+	diags.AddError(
+		common.ErrorReadHeader,
+		"Error during updating Tenant Settings.",
```
catalina-gav: Maybe also add the settingName for which the error occurred in the message

internal/services/tenantsetting/resource_tenant_setting.go
```diff
@@ -0,0 +1,265 @@
...
+
+	if resp.Diagnostics.Append(req.State.Get(ctx, &state)...); resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !state.DeleteBehaviour.IsNull() && !state.DeleteBehaviour.IsUnknown() && state.DeleteBehaviour.ValueString() == string(Disable) {
+		var reqUpdate requestUpdateTenantSettings
+		disabled := false
+		reqUpdate.UpdateTenantSettingRequest.Enabled = &disabled
+
```
catalina-gav, line 228: add timeout handling before API call
```suggestion
ctx, cancel := context.WithTimeout(ctx, timeout)
	defer cancel()
```

internal/services/tenantsettings/resource_tenant_settings.go
```diff
@@ -0,0 +1,265 @@
...
+
+		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
+	}
+
+	tflog.Debug(ctx, "DELETE", map[string]any{
+		"action": "end",
+	})
+}
+
+func (d *resourceTenantSettings) get(ctx context.Context, model *resourceTenantSettingsModel) diag.Diagnostics {
```
catalina-gav: r *resourceTenantSettings instead of d *resourceTenantSettings

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+				Resource: &schemaR.StringAttribute{
+					Computed: true,
+				},
+				DataSource: &schemaD.StringAttribute{
+					Computed: true,
+				},
+			},
+			"delete_behaviour": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "Indicates whether the tenant setting is disabled when deleted. False - The tenant setting is not disabled when deleted. True - The tenant setting is disabled when deleted.",
```
catalina-gav: Is this description from a previous delete approach? It should describe the No change/Disabled options

internal/services/tenantsetting/schema.go
```diff
@@ -0,0 +1,282 @@
...
+			"delete_behaviour": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "Indicates whether the tenant setting is disabled when deleted. False - The tenant setting is not disabled when deleted. True - The tenant setting is disabled when deleted.",
+					Validators: []validator.String{
+						stringvalidator.OneOf(utils.ConvertEnumsToStringSlices(PossibleDeleteBehaviourValues(), true)...),
+					},
+				},
+				Resource: &schemaR.StringAttribute{
+					Computed: true,
+					Optional: true,
```
catalina-gav, line 68: you could also add the default option here: Default: stringdefault.StaticString(string(NoChange)),

internal/services/tenantsettings/fake_test.go
```diff
@@ -0,0 +1,126 @@
...
+	return tenantSettings
+}
+
+func fakeTestUpsert(entity fabadmin.TenantSetting) {
+	fakeTenantSettingsStore[*entity.SettingName] = entity
+}
+
+func fakeUpdateTenantSettings(
+	exampleResp fabadmin.TenantSetting,
+) func(ctx context.Context, tenantSettingName string, updateTenantSettingRequest fabadmin.UpdateTenantSettingRequest, options *fabadmin.TenantsClientUpdateTenantSettingOptions) (resp azfake.Responder[fabadmin.TenantsClientUpdateTenantSettingResponse], errResp azfake.ErrorResponder) {
```
catalina-gav: I think you could directly use the updateTenantSettingRequest  to form the updated tenant settings body to update in the store and remove the exampleResp

internal/services/tenantsetting/resource_tenant_setting_test.go
CatalinDabuleanu: Create doesnt exist for this artifact. You can only update a tenant setting. This test updates the setting from enabled to disabled and adds a graph_id and name for the setting.

Delete operation reverts the changes added, but because `delete_behaviour` is set to `Disable`, after Delete, the setting is kept to disabled (or updated to disabled) 

This is the design we discussed with @OrBaubergMicrosoft and @DariuszPorowski

catalina-gav: Thank you for the reply, however I've seen that there is no create API endpoint, but there is a  create method in the provider and it is called in this test case. This will only test the update flow in the provider's Create method. What I meant is to add a test case after this one to also test the update flow within the provider's Update method

internal/services/tenantsettings/resource_tenant_settings.go
```diff
@@ -0,0 +1,273 @@
...
+	if resp.Diagnostics.Append(req.State.Get(ctx, &state)...); resp.Diagnostics.HasError() {
+		return
+	}
+
+	timeout, diags := state.Timeouts.Delete(ctx, r.pConfigData.Timeout)
+	if resp.Diagnostics.Append(diags...); resp.Diagnostics.HasError() {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, timeout)
```
catalina-gav: you can move these 2 lines inside if block, since only there we do an API call

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,284 @@
...
+						stringvalidator.OneOf(utils.ConvertEnumsToStringSlices(PossibleDeleteBehaviourValues(), true)...),
+					},
+				},
+				Resource: &schemaR.StringAttribute{
+					Computed: true,
+					Optional: true,
+					Default:  stringdefault.StaticString(string(NoChange)),
+				},
+				DataSource: &schemaD.StringAttribute{
+					Computed: true,
```
catalina-gav: Should this be computed for data sources? since it s not coming from the API?

internal/services/tenantsetting/schema.go
```diff
@@ -0,0 +1,284 @@
...
+					Computed: true,
+				},
+				DataSource: &schemaD.SetNestedAttribute{
+					Computed: true,
+				},
+				Attributes: superschema.Attributes{
+					"graph_id": superschema.StringAttribute{
+						Common: &schemaR.StringAttribute{
+							MarkdownDescription: "The graph ID of the security group.",
```
catalina-gav, line 156:
```suggestion
							MarkdownDescription: "The graph ID of the security group.",
                            CustomType:          customtypes.UUIDType{},
```

internal/services/tenantsettings/data_tenant_setting.go
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2026 Microsoft Corporation
```
catalina-gav: I see that in all the other files outside tenant_settings, the headers are like this:
// Copyright Microsoft Corporation 2026
// SPDX-License-Identifier: MPL-2.0

internal/services/tenantsetting/base.go
```diff
@@ -0,0 +1,16 @@
+// Copyright Microsoft Corporation 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package tenantsettings
+
```
badeamarjieh, line 5: please rename the package name to `tenantsetting` as well as the folder name to `tenantsetting`

internal/services/tenantsetting/models.go
```diff
@@ -0,0 +1,289 @@
...
+func (to *baseTenantSettingsModel) set(ctx context.Context, from fabadmin.TenantSetting) diag.Diagnostics {
+	to.SettingName = types.StringPointerValue(from.SettingName)
+	to.TenantSettingGroup = types.StringPointerValue(from.TenantSettingGroup)
+	to.Title = types.StringPointerValue(from.Title)
+	to.CanSpecifySecurityGroups = types.BoolPointerValue(from.CanSpecifySecurityGroups)
+	to.Enabled = types.BoolPointerValue(from.Enabled)
+	to.DelegateToCapacity = types.BoolPointerValue(from.DelegateToCapacity)
+	to.DelegateToDomain = types.BoolPointerValue(from.DelegateToDomain)
+	to.DelegateToWorkspace = types.BoolPointerValue(from.DelegateToWorkspace)
+
```
badeamarjieh, line 47: always set the properties of the type `supertypes.SetNestedObjectValueOf` as well. like this

```
enabledSecurityGroups := supertypes.NewSingleNestedObjectValueOfNull[tenantSettingsSecurityGroup](ctx)
to.EnabledSecurityGroups = enabledSecurityGroups
```


and then you override the `enabledSecurityGroups` and `to.EnabledSecurityGroups` if `from.EnabledSecurityGroups != nil` (similar to what you have right now)

badeamarjieh: you can look at models.go of `gateway` resource/data-source for reference

badeamarjieh: and then remove the else in the if else below (for all properties of type ) `supertypes.SetNestedObjectValueOf`

internal/services/tenantsettings/models.go
```diff
@@ -0,0 +1,289 @@
...
+DATA-SOURCE (LIST)
+*/
+
+type dataSourceTenantSettingsModel struct {
+	Values supertypes.SetNestedObjectValueOf[baseTenantSettingsModel] `tfsdk:"values"`
+
+	Timeouts timeoutsD.Value `tfsdk:"timeouts"`
+}
+
+func (to *dataSourceTenantSettingsModel) set(ctx context.Context, from []fabadmin.TenantSetting) diag.Diagnostics {
```
badeamarjieh:
```suggestion
func (to *dataSourceTenantSettingsModel) setValues(ctx context.Context, from []fabadmin.TenantSetting) diag.Diagnostics {
```

internal/services/tenantsettings/models.go
```diff
@@ -0,0 +1,289 @@
...
+	slice := make([]*baseTenantSettingsModel, 0, len(from))
+	for _, ts := range from {
```
badeamarjieh:
```suggestion
	for _, entity := range from {
```

internal/services/tenantsettings/models.go
```diff
@@ -0,0 +1,289 @@
...
+		tenantSetting.set(ctx, ts)
+		slice = append(slice, &tenantSetting)
+	}
+
+	if diags := to.Values.Set(ctx, slice); diags.HasError() {
+		return diags
+	}
+
+	return nil
+}
```
badeamarjieh:
```suggestion
    slice := make([]*baseTenantSettingsModel, 0, len(from))

	for _, entity := range from {
		var entityModel baseTenantSettingsModel

		if diags := entityModel.set(ctx, entity); diags.HasError() {
			return diags
		}

		slice = append(slice, &entityModel)
	}

	return to.Values.Set(ctx, slice)
```

badeamarjieh: 1. you need to check if there were errors upon entityModel set and not in the end - catch early and throw.
2. no need for the last if because when you do `return to.Values.Set(ctx, slice)` if it has diags it will return them directly, otherwise it will return nil, so it is exactly like your current code logic

internal/services/tenantsetting/models.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	return nil
+}
+
+/*
+HELPER MODELS
+*/
+
+type tenantSettingsSecurityGroup struct {
```
badeamarjieh, line 216:
```suggestion
type TenantSettingSecurityGroup struct {
```

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+		Attributes: map[string]superschema.Attribute{
+			"setting_name": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "The name of the tenant setting.",
+				},
+				Resource: &schemaR.StringAttribute{
+					Required: true,
+				},
+				DataSource: &schemaD.StringAttribute{
+					Optional: true,
```
badeamarjieh: `!isList`

if its list then optional = true, if its not list (single tenant setting) then optional = false

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+					Computed: true,
+				},
+			},
+			"tenant_setting_group": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "Tenant setting group name.",
+				},
+				Resource: &schemaR.StringAttribute{
+					Computed: true,
```
badeamarjieh: move this to the common and remove Resource and Datasource - this is a common configuration for both of them so it should be under `Common`

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+				DataSource: &schemaD.StringAttribute{
+					Computed: true,
+				},
+			},
+			"title": superschema.StringAttribute{
+				Common: &schemaR.StringAttribute{
+					MarkdownDescription: "The title of the tenant setting.",
+				},
+				Resource: &schemaR.StringAttribute{
+					Computed: true,
```
badeamarjieh: same here

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+				},
+				DataSource: &schemaD.StringAttribute{
+					Computed: true,
+				},
+			},
+			"delete_behaviour": superschema.StringAttribute{
+				Resource: &schemaR.StringAttribute{
+					MarkdownDescription: "Indicates whether the tenant setting is disabled when deleted. NoChange - The tenant setting is not disabled when deleted. Disable - The tenant setting is disabled when deleted.",
```
badeamarjieh: Specifies how the tenant setting is handled when the resource is deleted. `NoChange` - Keeps the current setting unchanged on resource deletion. If the setting is enabled it remains enabled; if disabled it remains disabled. `Disable` - Disables the tenant setting when the resource is deleted.

internal/services/tenantsetting/schema.go
```diff
@@ -0,0 +1,282 @@
...
+					Computed: true,
+				},
+			},
+			"delete_behaviour": superschema.StringAttribute{
+				Resource: &schemaR.StringAttribute{
+					MarkdownDescription: "Indicates whether the tenant setting is disabled when deleted. NoChange - The tenant setting is not disabled when deleted. Disable - The tenant setting is disabled when deleted.",
+					Validators: []validator.String{
+						stringvalidator.OneOf(utils.ConvertEnumsToStringSlices(PossibleDeleteBehaviourValues(), true)...),
+					},
+					Computed: true,
```
badeamarjieh, line 67: remove. its not computed as its not related to the API response. its always known at plan time depending on the value and existence of this proeperty in the HCL config

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+			"can_specify_security_groups": superschema.BoolAttribute{
+				Common: &schemaR.BoolAttribute{
+					MarkdownDescription: "Indicates if the tenant setting is enabled for a security group. False - The tenant setting is enabled for the entire organization. True - The tenant setting is enabled for security groups.",
+				},
+				Resource: &schemaR.BoolAttribute{
+					Computed: true,
+				},
+				DataSource: &schemaD.BoolAttribute{
+					Computed: true,
+				},
```
badeamarjieh: put under `Common` like above

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+			},
+			"delegate_to_capacity": superschema.BoolAttribute{
+				Common: &schemaR.BoolAttribute{
+					MarkdownDescription: "Indicates whether the tenant setting can be delegated to a capacity admin. False - Capacity admin cannot override the tenant setting. True - Capacity admin can override the tenant setting.",
+				},
+				Resource: &schemaR.BoolAttribute{
+					Optional: true,
+					Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+				},
+			},
+			"delegate_to_domain": superschema.BoolAttribute{
+				Common: &schemaR.BoolAttribute{
+					MarkdownDescription: "Indicates whether the tenant setting can be delegated to a domain admin. False - Domain admin cannot override the tenant setting. True - Domain admin can override the tenant setting.",
+				},
+				Resource: &schemaR.BoolAttribute{
+					Optional: true,
+					Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+				},
+			},
+			"delegate_to_workspace": superschema.BoolAttribute{
+				Common: &schemaR.BoolAttribute{
+					MarkdownDescription: "Indicates whether the tenant setting can be delegated to a workspace admin. False - Workspace admin cannot override the tenant setting. True - Workspace admin can override the tenant setting.",
+				},
+				Resource: &schemaR.BoolAttribute{
+					Optional: true,
+					Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+				},
+			},
+			"enabled_security_groups": superschema.SuperSetNestedAttributeOf[tenantSettingsSecurityGroup]{
+				Common: &schemaR.SetNestedAttribute{
+					MarkdownDescription: "A list of enabled security groups.",
+				},
+				Resource: &schemaR.SetNestedAttribute{
+					Optional: true,
+					Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsetting/schema.go
```diff
@@ -0,0 +1,282 @@
...
+							MarkdownDescription: "The name of the security group.",
+						},
+						Resource: &schemaR.StringAttribute{
+							Optional: true,
+							Computed: true,
+						},
+						DataSource: &schemaD.StringAttribute{
+							Computed: true,
+						},
+					},
```
badeamarjieh, line 142: I believe at least one of them should be required no? how does this work?
maybe you can pass either one of them but not both together? if so it needs to be reflected in the schema here

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+					},
+				},
+			},
+			"excluded_security_groups": superschema.SuperSetNestedAttributeOf[tenantSettingsSecurityGroup]{
+				Common: &schemaR.SetNestedAttribute{
+					MarkdownDescription: "A list of excluded security groups.",
+				},
+				Resource: &schemaR.SetNestedAttribute{
+					Optional: true,
+					Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsetting/schema.go
```diff
@@ -0,0 +1,282 @@
...
```
badeamarjieh, line 171: same comment here

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+					},
+				},
+			},
+			"properties": superschema.SuperSetNestedAttributeOf[tenantSettingsProperty]{
+				Common: &schemaR.SetNestedAttribute{
+					MarkdownDescription: "Tenant setting properties.",
+				},
+				Resource: &schemaR.SetNestedAttribute{
+					Optional: true,
+					Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+				},
+				Attributes: superschema.Attributes{
+					"name": superschema.StringAttribute{
+						Common: &schemaR.StringAttribute{
+							MarkdownDescription: "The name of the property.",
+						},
+						Resource: &schemaR.StringAttribute{
+							Optional: true,
+							Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsettings/schema.go
```diff
@@ -0,0 +1,282 @@
...
+						},
+					},
+					"type": superschema.StringAttribute{
+						Common: &schemaR.StringAttribute{
+							MarkdownDescription: "The type of the property.",
+						},
+						Resource: &schemaR.StringAttribute{
+							Optional: true,
+							Computed: true,
```
badeamarjieh: put under `Common` like above

internal/services/tenantsetting/models.go
```diff
@@ -0,0 +1,289 @@
...
+	return nil
+}
+
+func (to *baseTenantSettingsModel) setUpdate(ctx context.Context, from []fabadmin.TenantSetting) diag.Diagnostics {
+	for _, entity := range from {
+		if entity.SettingName != nil && *entity.SettingName == to.SettingName.ValueString() {
+			if diags := to.set(ctx, entity); diags.HasError() {
+				return diags
+			}
+		}
```
badeamarjieh, line 199: `return nil`

to exit the loop if you found it already

internal/services/tenantsettings/models.go
```diff
@@ -0,0 +1,289 @@
...
+func (to *baseTenantSettingsModel) setUpdate(ctx context.Context, from []fabadmin.TenantSetting) diag.Diagnostics {
+	for _, entity := range from {
+		if entity.SettingName != nil && *entity.SettingName == to.SettingName.ValueString() {
+			if diags := to.set(ctx, entity); diags.HasError() {
+				return diags
+			}
+		}
+	}
+
+	return nil
```
badeamarjieh: if you get here it means you did not find the setting, which is not expected, but still it needs to be handled this way so I'd suggest:

```
	diags.AddError(
		common.ErrorReadHeader,
		"Unable to find tenant setting with 'settingName': "+*to.SettingName.ValueString(),
	)

	return diags
```

internal/services/tenantsettings/resource_tenant_settings.go
```diff
@@ -0,0 +1,273 @@
...
+		respUpdate, err := r.client.UpdateTenantSetting(ctx, state.SettingName.ValueString(), reqUpdate.UpdateTenantSettingRequest, nil)
+		if resp.Diagnostics.Append(utils.GetDiagsFromError(ctx, err, utils.OperationUpdate, nil)...); resp.Diagnostics.HasError() {
+			return
+		}
+
+		if resp.Diagnostics.Append(state.setUpdate(ctx, respUpdate.TenantSettings)...); resp.Diagnostics.HasError() {
+			return
+		}
+
+		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
```
badeamarjieh: redundant? the resource will be deleted from the state anyway

badeamarjieh (changes requested):

internal/services/tenantsetting/schema.go
CatalinDabuleanu: Due to the fact that delete_behaviour has attribute 'Default', docs generation fails. Computed is required in this case

image

PR #779, opened 2025-12-18
feat(wellknown): add support for Maps in Set-Wellknown script

catalina-gav:
# 📥 Pull Request

Add support for Maps in Set-Wellknown script


internal/testhelp/fixtures/map/map.json.tmpl
```diff
@@ -0,0 +1,24 @@
...
+      "compass": true,
+      "scale": true,
+      "traffic": true
+    },
+    "backgroundColor": null,
+    "theme": null
+  },
+  "dataSources": {
+    "lakehouses": [{{ .LAKEHOUSES | default "" }}],
+    "kqlDataBases": [
```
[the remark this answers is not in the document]
catalina-gav: this is taken from documentation

internal/testhelp/fixtures/map/map.json.tmpl
CatalinDabuleanu, line 1: we use singular directory names for fixtures. rename this as "map" instead of "maps"

internal/testhelp/fixtures/maps/map.json.tmpl
```diff
@@ -0,0 +1,24 @@
...
+      "pitch": true,
+      "compass": true,
+      "scale": true,
+      "traffic": true
+    },
+    "backgroundColor": null,
+    "theme": null
+  },
+  "dataSources": {
+    "lakehouses": [{{ .LAKEHOUSES | default "" }}],
```
OrBaubergMicrosoft: what's the meaning of this part?

catalina-gav: The LAKEHOUSES token replaces and array that accepts one or more JSON objects representing a lakehouse with the corresponding attributes: workspaceId and artifactId. I used an array placeholder instead of having a single json object because it s more accurate to the actual definition and this way i could even test the resource with multiple artifacts if i wanted to
https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions/map-definition#mapdetails-example
image

PR #806, opened 2026-02-02
fix(domain): added informations about Service Principal permissions o…

CatalinDabuleanu:
…n fabric_domain

# 📥 Pull Request

Added informations about creating a domain using Service Principal


badeamarjieh (changes requested):
So we could split it into 2 notes:
1. The first one is whether the resource/data-source uses underlying Admin APIs. if that's true, we should add a note saying `The caller must be a Fabric administrator`.
2. The second one is if the resource/data-source (1) uses underlying Admin APIs, and (2) supports service principal. In that case, we should add another note to link to this documentation https://learn.microsoft.com/en-us/fabric/admin/enable-service-principal-admin-apis#enable-service-principal-authentication

so I would suggest adding one property to the base.go which specifies if the resource/data-source requires the caller to be fabric Admin. e.g., `IsAdmin`, and add 2 notes (if needed), one based on `IsAdmin` and the other based on `IsAdmin` and `IsSPNSupported`. or maybe you can use the existing note for SP support to add text to it, or if you have other suggestions

PR #817, opened 2026-02-10
feat(spark_environment_settings): migrate Resource/Data-Source to Use GA APIs

### LINKED ISSUE #809: [FEAT] Migrate sprark_environment_settings Resource/Data-Source to Use GA APIs
### 🚀 Feature description

Migrate sprark_environment_settings Resource/Data-Source to Use GA APIs

### 🔈 Motivation

Preview APIs will be deprecated by March 1st

### 🛰 Alternatives

_No response_

### 🚧 Potential Configuration / Desired Solution

_No response_

### 📎 Additional context

_No response_

### ☑️ Acceptance Criteria

_No response_

### 🔰 Code of Conduct

### LINKED ISSUE #834: [BUG] Removing or omitting a spark property in fabric_spark_environment_settings will result in a state drift
### 🐛 What happened?

In a fabric_spark_environment resource, after having set a spark property, we want to omit it or set it to null in order to remove it from spark_properties. However, trying to do so will result in a state drift

It also results in a state drift if a spark property has been added outside the TF config 

### 🔬 How to reproduce?

run the sample code, then remove one of the spark properties and run again

### 🏗️ Code Sample / Log

```
resource "fabric_spark_environment_settings" "example" {
  workspace_id       = ""
  environment_id     = ""
  publication_status = "Staging"

  driver_cores  = 4
  driver_memory = "28g"

  executor_cores  = 4
  executor_memory = "28g"

  runtime_version = "1.2"
  spark_properties = {
    "spark.acls.enable" = "true"
  }
}
```

### 📷 Screenshots

Image
Image

### 📈 Expected behavior

When removing or omitting a spark_property key value pair it should get deleted. Also TF config spark_properties should override any existing spark_properties 

1.7.0

1.14.6.

### 🌌 Environment (OS)

Linux

### 📎 Additional context

_No response_

### 🔰 Code of Conduct

catalina-gav:
# 📥 Pull Request

Close #809 
Close #834


internal/services/sparkenvsettings/resource_spark_environment_settings.go
```diff
@@ -277,67 +276,55 @@ func (r *resourceSparkEnvironmentSettings) Delete(ctx context.Context, _ resourc
...
 				return diags
-
-			case strings.ToLower((string)(fabenvironment.PublishStateSuccess)):
-				return nil
-
-			default:
-				tflog.Info(ctx, "Environment provisioning in progress, waiting 30 seconds before retrying")
-				time.Sleep(30 * time.Second) // lintignore:R018
 			}
 		}
```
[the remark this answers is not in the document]
catalina-gav: PublishEnvironment (the GA wrapper) internally calls BeginPublishEnvironment and .Sync(ctx), which polls the LRO to completion before returning. The SDK handles the polling for us, so manual polling is no longer needed. The state checks for Failed/Cancelled are kept as a defensive safety net.

internal/services/sparkenvsettings/resource_spark_environment_settings_test.go
```diff
@@ -63,7 +63,52 @@ func TestAcc_SparkEnvironmentSettingsResource_CRUD(t *testing.T) {
 				resource.TestCheckResourceAttr(testResourceItemFQN, "driver_cores", "8"),
 			),
 		},
-		// Update and Read (Spark properties)
+	},
+	))
+}
+
```
CatalinDabuleanu, line 282: please add unit tests for this resource too

internal/services/sparkenvsettings/models.go
```diff
@@ -143,15 +119,15 @@ func (to *requestUpdateSparkEnvironmentSettings) set(ctx context.Context, from r
 	}
 
 	if !from.DriverMemory.IsNull() && !from.DriverMemory.IsUnknown() {
-		to.DriverMemory = from.DriverMemory.ValueStringPointer()
+		to.DriverMemory = (*fabenvironment.CustomPoolMemory)(from.DriverMemory.ValueStringPointer())
```
OrBaubergMicrosoft, line 125: why was this change needed? did they add a new enum?

internal/services/sparkenvsettings/resource_spark_environment_settings.go
```diff
@@ -221,19 +231,30 @@ func (r *resourceSparkEnvironmentSettings) Update(ctx context.Context, req resou
...
+	{
+		respCurrent, err := r.stagingClient.GetSparkCompute(ctx, plan.WorkspaceID.ValueString(), plan.EnvironmentID.ValueString(), false, nil)
+		if resp.Diagnostics.Append(utils.GetDiagsFromError(ctx, err, utils.OperationUpdate, nil)...); resp.Diagnostics.HasError() {
+			return
+		}
+
+		if reqUpdate.SparkProperties != nil || len(respCurrent.SparkCompute.SparkProperties) > 0 {
+			reqUpdate.SparkProperties = diffSparkProperties(reqUpdate.SparkProperties, respCurrent.SparkCompute.SparkProperties)
+		}
+	}
```
OrBaubergMicrosoft: not sure I understand this one
why do we need this logic?

internal/services/sparkenvsettings/models.go
catalina-gav: Yes, from string:
image
to CustomPoolMemory
image
Same was done for executorMemory

internal/services/sparkenvsettings/resource_spark_environment_settings.go
catalina-gav: https://learn.microsoft.com/en-us/rest/api/fabric/environment/staging/update-spark-compute?tabs=HTTP
UpdateSparkCompute uses PATCH semantics: it merges sparkProperties instead of replacing them.
That means properties added outside Terraform remain unless explicitly set to null for removal.
In the provider, if users update some properties while others were added in the UI for eg, the API response includes both sets, which will cause Terraform state drift.

catalina-gav: This solution builds a request body that keeps Terraform-managed properties and sets externally added ones to null, effectively mimicking full override behavior.

internal/services/sparkenvsettings/models.go
```diff
@@ -214,15 +190,18 @@ func (to *requestUpdateSparkEnvironmentSettings) set(ctx context.Context, from r
...
 
-		for k, v := range sparkProperties {
-			if !v.IsNull() && !v.IsUnknown() {
-				sparkPropertiesMap[k] = v.ValueString()
-			}
+		for _, prop := range sparkProperties {
+			var reqProp fabenvironment.SparkProperty
+
+			reqProp.Key = prop.Key.ValueStringPointer()
+			reqProp.Value = prop.Value.ValueStringPointer()
```
OrBaubergMicrosoft: why not reqProp.set(prop)?

internal/services/sparkenvsettings/models.go
```diff
@@ -255,3 +234,42 @@ func (to *instancePoolPropertiesModel) set(from fabenvironment.InstancePool) {
...
+
+// diffSparkProperties merges planned spark properties with current ones,
+// adding null-value entries for any current keys not present in the plan.
+// This ensures the API deletes properties that were removed from config.
+func diffSparkProperties(planned, current []fabenvironment.SparkProperty) []fabenvironment.SparkProperty {
+	plannedKeys := make(map[string]struct{})
+
+	for _, p := range planned {
+		if p.Key != nil {
+			plannedKeys[*p.Key] = struct{}{}
```
OrBaubergMicrosoft: what is that? I am not familiar

internal/services/sparkenvsettings/resource_spark_environment_settings.go
```diff
@@ -277,67 +298,55 @@ func (r *resourceSparkEnvironmentSettings) Delete(ctx context.Context, _ resourc
...
-				return diags
-			}
-
-			if respPublish.PublishDetails == nil || respPublish.PublishDetails.State == nil {
-				tflog.Info(ctx, "Environment publishing not done, waiting 30 seconds before retrying")
-				time.Sleep(30 * time.Second) // lintignore:R018
+		respPublish, err := r.itemsClient.PublishEnvironment(ctx, model.WorkspaceID.ValueString(), model.EnvironmentID.ValueString(), false, nil)
+		if diags := utils.GetDiagsFromError(ctx, err, utils.OperationCreate, nil); diags.HasError() {
+			return diags
+		}
```
OrBaubergMicrosoft, line 327: what about this logic? why is it no longer needed?

catalina-gav: This was specificied in the deprecation notice https://learn.microsoft.com/en-us/rest/api/fabric/environment/items/publish-environment(beta)?tabs=HTTP#deprecation-notice

> The following incompatible changes were introduced in the Release version:
> 
> Response is following [long running operations (LRO)](https://learn.microsoft.com/en-us/rest/api/fabric/articles/long-running-operation) pattern and HTTP response code 202 may be returned.

Since this is now a long-running operation (LRO), polling/retry handling is performed by the Fabric SDK, so we no longer need manual provider-side polling logic.

internal/services/sparkenvsettings/models.go
catalina-gav: we have the func (to *sparkPropertyModel) set(from fabenvironment.SparkProperty), so it maps SDK → model.
Here we need model → SDK, which doesn’t exist.

internal/services/sparkenvsettings/models.go
catalina-gav: transformed the map into a set for faster filtering, but will change this method to be easier to read

docs/resources/spark_environment_settings.md
```diff
@@ -44,11 +41,13 @@ resource "fabric_spark_environment_settings" "example" {
    */
   }
 
-  spark_properties = {
-    /*
+  spark_properties = [
```
OrBaubergMicrosoft: please add an example with spark properties

internal/services/sparkenvsettings/schema.go
```diff
@@ -189,28 +186,49 @@ func itemSchema() superschema.Schema { //nolint:maintidx
 					Computed: true,
 				},
 			},
-			"spark_properties": superschema.SuperMapAttribute{
-				Common: &schemaR.MapAttribute{
-					MarkdownDescription: "A map of key/value pairs of Spark properties.",
-					CustomType:          supertypes.MapTypeOf[types.String]{MapType: types.MapType{ElemType: types.StringType}},
-					ElementType:         types.StringType,
+			"spark_properties": superschema.SuperListNestedAttributeOf[sparkPropertyModel]{
```
OrBaubergMicrosoft: this is aligning with the API , but it will break our customers, 
can you keep the old representation and schema and transform it in out BE to KV list?

internal/services/sparkenvsettings/schema.go
```diff
@@ -197,7 +194,6 @@ func itemSchema() superschema.Schema { //nolint:maintidx
 				},
 				Resource: &schemaR.MapAttribute{
 					Optional: true,
-					Computed: true,
```
OrBaubergMicrosoft, line 200: why was it removed?

PR #829, opened 2026-02-24
fix(lint): fix linting issues

CatalinDabuleanu:
# 📥 Pull Request

fix lint errors


internal/services/azuremap/base.go
```diff
@@ -1,7 +1,7 @@
 // Copyright Microsoft Corporation 2026
 // SPDX-License-Identifier: MPL-2.0
 
-package maps
+package azuremap
```
OrBaubergMicrosoft, line 4: change to fabricmap (in tests as well)

PR #870, opened 2026-03-22
feat(fabric_item): added sensitivity label to fabric item

CatalinDabuleanu:
# 📥 Pull Request

Implemented sensitivity label for fabric item


internal/pkg/fabricitem/resource_schema.go
```diff
@@ -205,6 +206,31 @@ func getResourceFabricItemBaseAttributes(
 			Optional:            true,
 			CustomType:          customtypes.UUIDType{},
 		},
+		"sensitivity_label_settings": schema.SingleNestedAttribute{
+			MarkdownDescription: "The sensitivity label settings. Once set, changing this value will force recreation of the resource.",
```
badeamarjieh: no need for this, once you set RequireReplace modifier the docs will automatically add a similar sentence
https://registry.terraform.io/providers/microsoft/fabric/latest/docs/resources/deployment_pipeline#:~:text=(Attributes%20List)-,(ForceNew),-The%20collection%20of

internal/pkg/fabricitem/resource_schema.go
```diff
@@ -205,6 +206,31 @@ func getResourceFabricItemBaseAttributes(
...
+			Optional:            true,
+			CustomType:          supertypes.NewSingleNestedObjectTypeOf[sensitivityLabelSettingsModel](ctx),
+			PlanModifiers: []planmodifier.Object{
+				objectplanmodifier.RequiresReplace(),
```
badeamarjieh, line 214: I am wondering if setting RequireReplace at the sensitivity_label_settings level, would trigger a replace if you change only one if its propreties (like label_id or the sensitivity_label_apply_strategy)

internal/pkg/fabricitem/resource_schema.go
```diff
@@ -205,6 +206,31 @@ func getResourceFabricItemBaseAttributes(
...
+				},
+				"sensitivity_label_apply_strategy": schema.StringAttribute{
+					MarkdownDescription: fmt.Sprintf(
+						"The strategy for applying the sensitivity label. Possible values: %s.",
+						utils.ConvertStringSlicesToString(fabcore.PossibleSensitivityLabelApplyStrategyValues(), true, true),
+					),
+					Optional: true,
+					Validators: []validator.String{
+						stringvalidator.OneOf(utils.ConvertEnumsToStringSlices(fabcore.PossibleSensitivityLabelApplyStrategyValues(), true)...),
+					},
```
badeamarjieh, line 233: add default value to be ApplyOrFail (as per the API docs).

internal/pkg/fabricitem/models.go
```diff
@@ -4,46 +4,95 @@
...
 	to.Description = types.StringPointerValue(from.Description)
 	to.FolderID = customtypes.NewUUIDPointerValue(from.FolderID)
+
+	return to.setSensitivityLabelSettings(ctx, from.SensitivityLabel)
+}
+
+func (to *fabricItemModel) setSensitivityLabelSettings(ctx context.Context, sl *fabcore.SensitivityLabel) diag.Diagnostics {
+	slSettings := supertypes.NewSingleNestedObjectValueOfNull[sensitivityLabelSettingsModel](ctx)
+
+	if sl != nil && sl.ID != nil {
```
badeamarjieh: is this needed? ID is required in the schema

internal/pkg/fabricitem/models.go
```diff
@@ -4,46 +4,95 @@
...
+		slModel := &sensitivityLabelSettingsModel{
```
badeamarjieh: slSettingsModel

internal/pkg/fabricitem/models_resource_item.go
```diff
@@ -43,6 +43,28 @@ func (to *requestCreateFabricItem) setFolderID(v customtypes.UUID) {
...
+		slModel, diags := v.Get(ctx)
+		if diags.HasError() {
+			return diags
+		}
+
+		settings := &fabcore.SensitivityLabelSettings{
+			LabelID: slModel.LabelID.ValueStringPointer(),
+		}
+
+		if !slModel.SensitivityLabelApplyStrategy.IsNull() && !slModel.SensitivityLabelApplyStrategy.IsUnknown() {
```
badeamarjieh: this is probably not needed as SensitivityLabelApplyStrategy should have a Default value.

internal/pkg/fabricitem/models_resource_item.go
```diff
@@ -43,6 +43,28 @@ func (to *requestCreateFabricItem) setFolderID(v customtypes.UUID) {
...
+			strategy := fabcore.SensitivityLabelApplyStrategy(slModel.SensitivityLabelApplyStrategy.ValueString())
```
badeamarjieh: please wrap this with parenthesis

internal/pkg/fabricitem/models.go
```diff
@@ -4,46 +4,95 @@
...
-	DisplayName types.String     `tfsdk:"display_name"`
-	Description types.String     `tfsdk:"description"`
-	FolderID    customtypes.UUID `tfsdk:"folder_id"`
+	WorkspaceID              customtypes.UUID                                                    `tfsdk:"workspace_id"`
+	ID                       customtypes.UUID                                                    `tfsdk:"id"`
+	DisplayName              types.String                                                        `tfsdk:"display_name"`
+	Description              types.String                                                        `tfsdk:"description"`
+	FolderID                 customtypes.UUID                                                    `tfsdk:"folder_id"`
+	SensitivityLabelSettings supertypes.SingleNestedObjectValueOf[sensitivityLabelSettingsModel] `tfsdk:"sensitivity_label_settings"`
 }
```
badeamarjieh: the fabricItemModel is the base model for both resource and datasource model. However, the sensitivity_label_settings schema will be different for resource and data-source:
In resource it will the contract `SensitivityLabelSettings` (which includes  both `label_id` and `sensitivity_label_apply_strategy`):https://learn.microsoft.com/en-us/rest/api/fabric/core/items/create-item?tabs=HTTP#:~:text=as%20its%20folder.-,sensitivityLabelSettings,The%20sensitivity%20label%20settings%20for%20the%20item.,-Responses

Whereas in data-source it will have the contract `SensitivityLabel` (which includes only `label_id`): https://learn.microsoft.com/en-us/rest/api/fabric/core/items/get-item?tabs=HTTP#sensitivitylabel:~:text=The%20item%20ID.-,sensitivityLabel,The%20item%20sensitivity%20label.,-tags

Therefore, you should add different contracts not to the base resource and data-source

internal/pkg/fabricitem/models.go
```diff
@@ -63,6 +112,7 @@ func (to *FabricItemProperties[Titemprop]) Set(from any) {
 	to.DisplayName = getFieldStringValue(fromValue, "DisplayName")
 	to.Description = getFieldStringValue(fromValue, "Description")
 	to.FolderID = getFieldStringValue(fromValue, "FolderID")
+	to.SensitivityLabel = getFieldStructValue[fabcore.SensitivityLabel](fromValue, "SensitivityLabel")
 	to.Properties = getFieldStructValue[Titemprop](fromValue, "Properties")
```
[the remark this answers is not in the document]
CatalinDabuleanu: this is standard behavior for other fields too

badeamarjieh (changes requested):

internal/pkg/fabricitem/resource_schema.go
CatalinDabuleanu: Yes, it would. UpdateItemRequest does not expose any fields for sensitivity, so if we want to update anything, we need to recreate the whole object

internal/pkg/fabricitem/models_data_item_definition.go
```diff
@@ -18,12 +18,26 @@ import (
 type dataSourceFabricItemDefinitionModel struct {
 	fabricItemModel
 
+	SensitivityLabel supertypes.SingleNestedObjectValueOf[sensitivityLabelModel]                `tfsdk:"sensitivity_label"`
```
badeamarjieh: consider wrapping fabricItemModel with dataFabricItemModelBase (or the agreed upon naming - see other places) which adds the sensitivty label to it, and then simply replace the fabricItemModel to this new datasource fabric item model base.

same for the resource

badeamarjieh: this should also simply things as in line 29 - `to.fabricItemModel.set(from)` would change to `to.dataFabricItemModelBase.set(from)` which inside of it you set the sentivity label and reduce code duplication

internal/pkg/fabricitem/models.go
```diff
@@ -29,6 +56,31 @@ func (to *fabricItemModel) set(from fabcore.Item) {
...
+type dataSourceFabricItemBaseModel struct {
+	fabricItemModel
+
+	SensitivityLabel supertypes.SingleNestedObjectValueOf[sensitivityLabelModel] `tfsdk:"sensitivity_label"`
+}
+
+func (to *dataSourceFabricItemBaseModel) set(ctx context.Context, from fabcore.Item) diag.Diagnostics {
+	to.fabricItemModel.set(from)
+
+	sl, diags := newSensitivityLabelFromAPI(ctx, from.SensitivityLabel)
```
badeamarjieh: rename sl to sensitivityLabel, here and in all places

internal/pkg/fabricitem/models.go
```diff
@@ -4,15 +4,42 @@
...
+type sensitivityLabelSettingsModel struct {
+	LabelID                       customtypes.UUID `tfsdk:"label_id"`
+	SensitivityLabelApplyStrategy types.String     `tfsdk:"sensitivity_label_apply_strategy"`
+}
+
+type sensitivityLabelModel struct {
+	LabelID customtypes.UUID `tfsdk:"label_id"`
+}
+
+func newSensitivityLabelFromAPI(ctx context.Context, sl *fabcore.SensitivityLabel) (supertypes.SingleNestedObjectValueOf[sensitivityLabelModel], diag.Diagnostics) {
```
badeamarjieh: this be `set` method, just like fabricItemModel `set` method.

badeamarjieh: I am not sure about this pattern.

should be similar to this for exmaple:

```go
type publicKeyModel struct {
	Exponent types.String `tfsdk:"exponent"`
	Modulus  types.String `tfsdk:"modulus"`
}

func (to *publicKeyModel) set(from fabcore.PublicKey) {
	to.Exponent = types.StringPointerValue(from.Exponent)
	to.Modulus = types.StringPointerValue(from.Modulus)
}
```

where in the other function you do:
```go
...
	publicKey := supertypes.NewSingleNestedObjectValueOfNull[publicKeyModel](ctx)
    ...
    to.PublicKey = publicKey
    ...
    		if entity.PublicKey != nil {
			publicKeyModel := &publicKeyModel{}
			publicKeyModel.set(*entity.PublicKey)

			if diags := publicKey.Set(ctx, publicKeyModel); diags.HasError() {
				return diags
			}
		}

		to.PublicKey = publicKey
```

you can look at models.go of gateway

internal/pkg/fabricitem/models_data_items.go
```diff
@@ -15,17 +15,29 @@ import (
...
+	Timeouts    timeouts.Value                                                   `tfsdk:"timeouts"`
 }
 
 func (to *dataSourceFabricItemsModel) setValues(ctx context.Context, from []fabcore.Item) diag.Diagnostics {
-	slice := make([]*fabricItemModel, 0, len(from))
+	slice := make([]*dataSourceFabricItemBaseModel, 0, len(from))
 
 	for _, entity := range from {
-		var entityModel fabricItemModel
-		entityModel.set(entity)
```
badeamarjieh, line 28: you can keep it as is, dataSourceFabricItemBaseModel has `set` method which also sets the sensitivity label

badeamarjieh (approved):

badeamarjieh (approved):

internal/testhelp/fakes/fabric_item.go
```diff
@@ -42,6 +42,12 @@ func (o *operationsItem) CreateWithParentID(parentID string, data fabcore.Create
 	result.Description = data.Description
 	result.FolderID = data.FolderID
 
+	if data.SensitivityLabelSettings != nil {
+		result.SensitivityLabel = &fabcore.SensitivityLabel{
+			ID: data.SensitivityLabelSettings.LabelID,
```
OrBaubergMicrosoft, line 47: @badeamarjieh why only for fabric item fakes?

PR #947, opened 2026-04-25
feat(wellknown): add snowflake database to wellknown

CatalinDabuleanu:
# 📥 Pull Request

Add Snowflake Database to wellknown

### DIFF
--- a/tools/scripts/Set-WellKnown.ps1
+++ b/tools/scripts/Set-WellKnown.ps1
@@ -318,6 +318,9 @@ function Set-FabricItem {
     'SemanticModel' {
       $itemEndpoint = 'semanticModels'
     }
+    'SnowflakeDatabase' {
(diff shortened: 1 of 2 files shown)


tools/scripts/Set-WellKnown.ps1
```diff
@@ -1400,6 +1404,25 @@ $wellKnown['KQLDatabase'] = @{
   description = $kqlDatabase.description
 }
 
+# Create SnowflakeDatabase if not exists
```
catalina-gav, line 1407: I see that in the API there is an example with definition created without definition/creation payload, so you could simply create it without definition

CatalinDabuleanu: that is a doc bug. you cannot create a snowflake db without a definition/creation payload

catalina-gav (approved):

PR #980, opened 2026-05-13
fix(connection): added missing import functionality for fabric_conn…

### LINKED ISSUE #729: [Feature] add fabric_connection import state
### 🐛 What happened?

I tried to import 2 OAuth connections (which I re-implemented as terraform/OpenTofu code), that were created from a fabric pipeline, and despite [the docs](https://registry.terraform.io/providers/microsoft/fabric/latest/docs/resources/connection) stating that imports should be possible by their connection_id, my `tofu apply` command resulted in the error:

```shell
│ Error: Resource Import Not Implemented
│ 
│ This resource does not support import. Please contact the provider developer for additional information.
╵
```

### 🔬 How to reproduce?

Code samples will follow below for the terraform code... 
1. Setup the fabric provider with the `preview` setting set to `true` (i.e. connections are a preview feature)
2. Setup a Fabric pipeline with a copy-activity that uses a Fabric Lakehouse with parquet files as a source connection, and a Fabric Warehouse as the target connection (and create the connections through the pipeline editor UI).
3. Verify in the Fabric UI > Settings > Manage Connections and gateways that 2 connections have been created that are linked to the user account that authored the pipeline
4. Re-implement those examples using the `fabric_connection` resource, and add import blocks in the root `main.tf` that attempt to link the connections to the the `fabric_connection` resources.
5. Run `terraform apply`
6. Observe that terraform produces the error stated above, even though it remains implicit which resource it's referring too in the error message.

### 🏗️ Code Sample / Log

Provider setup (minimal example):

```hcl
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "=4.50.0"
    }

    fabric = {
      source  = "microsoft/fabric"
      version = "1.6.0"
    }

data "azurerm_client_config" "current" {}

provider "fabric" {
  tenant_id = data.azurerm_client_config.current.tenant_id
  use_cli   = true
  preview   = true
}
```

Main.tf with resources + imports:

```hcl
resource "fabric_connection" "lakehouse-oauth2-connection" {
  display_name       = "Lakehouse username"
  connectivity_type  = "ShareableCloud"
  connection_details = {
    creation_method = "Lakehouse.Contents"
    type = "Lakehouse"
  }
  credential_details = {
    credential_type = "OAuth2"
    connection_encryption = "NotEncrypted"
  }

  privacy_level = "Private"
}

resource "fabric_connection" "warehouse-oauth2-connection" {
  display_name       = "Warehouse username"
  connectivity_type  = "ShareableCloud"
  connection_details = {
    creation_method = "Fabric.Warehouse"
    type = "Warehouse"
  }
  credential_details = {
    credential_type = "OAuth2"
    connection_encryption = "NotEncrypted"
  }
}

import {
  id = "my-lakehouse-connection-id"
  to = fabric_connection.lakehouse-oauth2-connection
}

import {
  id = "my-warehouse-connection-id"
  to = fabric_connection.warehouse-oauth2-connection
}
```

In the above example, `username`, `my-lakehouse-connection-id`, and `my-warehouse-connection-id` should be replaced with the actual username (who authored the pipeline), and connection id's data factory created. Which can be found under the settings of each connection.

### 📷 Screenshots

The instructions on the Fabric provider docs page stating that imports should be possible:

Image

The resulting error that states imports are not implemented:

Image

### 📈 Expected behavior

I expected the Fabric connections would be imported as stated by the docs of that being possible.

1.6.0

OpenTofu 1.10.7

### 🌌 Environment (OS)

Windows

### 📎 Additional context

If there's any more info required, feel free to ask...

### 🔰 Code of Conduct

CatalinDabuleanu:
# 📥 Pull Request

Added missing import functionality to `fabric_connection`

### DIFF
--- a/internal/services/connection/models_resource_connection.go
+++ b/internal/services/connection/models_resource_connection.go
@@ -69,9 +69,13 @@ func (m rsConnectionDetailsModel) getParameters(ctx context.Context) (map[string
 	return nil, nil
 }
 
-func setRSConnectionDetails(from fabcore.ListConnectionDetails, to *rsConnectionDetailsModel) {
+func setRSConnectionDetails(ctx context.Context, from fabcore.ListConnectionDetails, to *rsConnectionDetailsModel) {
 	to.Path = types.StringPointerValue(from.Path)
 	to.Type = types.StringPointerValue(from.Type)
+
+	if to.Parameters.IsNull() && !to.Parameters.IsKnown() {
+		to.Parameters = supertypes.NewSetNestedObjectValueOfNull[connectionParametersModel](ctx)
+	}
 }
 
 type rsCredentialDetailsModel struct {
@@ -87,11 +91,31 @@ type rsCredentialDetailsModel struct {
 	SharedAccessSignatureCredentials supertypes.SingleNestedObjectValueOf[credentialsSharedAccessSignatureModel] `tfsdk:"shared_access_signature_credentials"`
 }
 
-func setRSCredentialDetails(from fabcore.ListCredentialDetails, to *rsCredentialDetailsModel) {
+func setRSCredentialDetails(ctx context.Context, from fabcore.ListCredentialDetails, to *rsCredentialDetailsModel) {
 	to.ConnectionEncryption = types.StringPointerValue((*string)(from.ConnectionEncryption))
 	to.SingleSignOnType = types.StringPointerValue((*string)(from.SingleSignOnType))
 	to.SkipTestConnection = types.BoolPointerValue(from.SkipTestConnection)
 	to.CredentialType = types.StringPointerValue((*string)(from.CredentialType))
+
+	if to.BasicCredentials.IsNull() && !to.BasicCredentials.IsKnown() {
+		to.BasicCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsBasicModel](ctx)
+	}
+
+	if to.KeyCredentials.IsNull() && !to.KeyCredentials.IsKnown() {
+		to.KeyCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsKeyModel](ctx)
+	}
+
+	if to.KeyPairCredentials.IsNull() && !to.KeyPairCredentials.IsKnown() {
+		to.KeyPairCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsKeyPairModel](ctx)
+	}
+
+	if to.ServicePrincipalCredentials.IsNull() && !to.ServicePrincipalCredentials.IsKnown() {
+		to.ServicePrincipalCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsServicePrincipalModel](ctx)
+	}
+
+	if to.SharedAccessSignatureCredentials.IsNull() && !to.SharedAccessSignatureCredentials.IsKnown() {
+		to.SharedAccessSignatureCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsSharedAccessSignatureModel](ctx)
+	}
 }
 
 func getKeyVaultSecretReference(ctx context.Context, kvRef supertypes.SingleNestedObjectValueOf[credentialsKeyVaultReferenceModel]) (*fabcore.KeyVaultSecretReference, diag.Diagnostics) {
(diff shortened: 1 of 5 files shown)


internal/services/connection/models_resource_connection.go
```diff
@@ -87,11 +91,31 @@ type rsCredentialDetailsModel struct {
...
+	if to.KeyPairCredentials.IsNull() && !to.KeyPairCredentials.IsKnown() {
+		to.KeyPairCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsKeyPairModel](ctx)
+	}
+
+	if to.ServicePrincipalCredentials.IsNull() && !to.ServicePrincipalCredentials.IsKnown() {
+		to.ServicePrincipalCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsServicePrincipalModel](ctx)
+	}
+
+	if to.SharedAccessSignatureCredentials.IsNull() && !to.SharedAccessSignatureCredentials.IsKnown() {
+		to.SharedAccessSignatureCredentials = supertypes.NewSingleNestedObjectValueOfNull[credentialsSharedAccessSignatureModel](ctx)
```
OrBaubergMicrosoft, line 117: why is this change needed? is it for exported connections?

internal/services/connection/resource_connection_test.go
```diff
@@ -1379,6 +1382,89 @@ func TestUnit_ConnectionResource_ModifyPlan_Validations(t *testing.T) {
...
+			Config:        testCase,
+			ImportStateId: fmt.Sprintf("%s/%s", *entity.ID, "test"),
+			ImportState:   true,
+			ExpectError:   regexp.MustCompile(customtypes.UUIDTypeErrorInvalidStringHeader),
+		},
+		// Import state testing
+		{
+			ResourceName:       testResourceItemFQN,
+			Config:             testCase,
+			ImportStateId:      *entity.ID,
```
OrBaubergMicrosoft, line 1450: please add .sh file, like we have in other items

internal/services/connection/resource_connection_test.go
```diff
@@ -1379,6 +1382,89 @@ func TestUnit_ConnectionResource_ModifyPlan_Validations(t *testing.T) {
...
+			ImportState:        true,
+			ImportStatePersist: true,
+			ImportStateCheck: func(is []*terraform.InstanceState) error {
+				if len(is) != 1 {
+					return errors.New("expected one instance state")
+				}
+
```
OrBaubergMicrosoft, line 1457: did you try this manually and checked the resulted state?
does the state has all needed data?

internal/services/connection/models_resource_connection.go
CatalinDabuleanu: Prior to having import, the Read operation would always be called after Create/Update which is fine

With import workflow is

1. user runs import
2. ImportState runs - it only sets the id attribute via ImportStatePassthroughID. Every other attribute in state is left at its Go zero value.
3. Terraform immediately calls Read to refresh state. The framework decodes the just-imported state into a resourceConnectionModel, but since only id exists, all nested supertypes fields are Go zero-value structs (supertypes.SingleNestedObjectValueOf[T]{}) - internally these have nil attribute-type metadata and are neither Null nor Known.
4. Read calls r.get - eventually setRSCredentialDetails(ctx, ..., to). The SDK GET response doesn't return credentials, so the code never assigns to to.BasicCredentials etc. Those fields remain zero-value.
5. The framework then serializes the model back to state via ToTerraformValue. Calling it on a zero-value supertypes.SingleNestedObjectValueOf[T] dereferences nil internals - panic.

internal/services/connection/resource_connection_test.go
CatalinDabuleanu: The file does exist https://github.com/microsoft/terraform-provider-fabric/blob/673fb74f1e6768e9fb5bcf7315a2f7a99fe58aa9/examples/resources/fabric_connection/import.sh

internal/services/connection/resource_connection_test.go
CatalinDabuleanu: yes, this worked as expected

OrBaubergMicrosoft (approved):

PR #1005, opened 2026-06-03
fix(definitions): fix api and provider item definitions inconsistencies

### LINKED ISSUE #1002: [BUG] API docs and definition inconsistencies
### 🐛 What happened?

Some onboarded items have definition rules and parts that do not match the current Fabric API documentation. Specifically:

- Certain definition parts documented by the API are missing from the provider's item configuration
- Some validation rules are too strict — rejecting file paths/parts that are valid and accepted by the API, causing unnecessary apply failures

### 🔬 How to reproduce?

_No response_

### 🏗️ Code Sample / Log

_No response_

### 📷 Screenshots

_No response_

### 📈 Expected behavior

The provider's definition rules and supported parts for each item should match the Fabric API documentation. If the API accepts a file path/part in the definition, the provider should accept it as well. If the API documents additional definition parts for an item, the provider should support them.

latest

latest

### 🌌 Environment (OS)

Linux

### 📎 Additional context

_No response_

### 🔰 Code of Conduct

catalina-gav:
# 📥 Pull Request

Close #1002


internal/services/digitaltwinbuilder/resource_digital_twin_builder.go
```diff
@@ -20,8 +25,19 @@ func NewResourceDigitalTwinBuilder() resource.Resource {
 		DescriptionMaxLength:  256,
 		DefinitionPathDocsURL: ItemDefinitionPathDocsURL,
 		DefinitionPathKeysValidator: []validator.Map{
-			mapvalidator.SizeAtMost(1),
-			mapvalidator.KeysAre(fabricitem.DefinitionPathKeysValidator(itemDefinitionFormats)...),
+			mapvalidator.SizeAtLeast(1),
+			mapvalidator.KeysAre(
+				fwvalidators.PatternsIfAttributeIsOneOf(
```
OrBaubergMicrosoft, line 30: is this the pattern across the provider?
default is the only definition format, seems a bit weird

OrBaubergMicrosoft: what about https://github.com/microsoft/terraform-provider-fabric/blob/d481355e9936874fb67945edd852cdae24211174/internal/services/lakehouse/resource_lakehouse.go#L104

internal/services/fabricmap/base.go
```diff
@@ -29,8 +29,11 @@ var ItemTypeInfo = tftypeinfo.TFTypeInfo{ //nolint:gochecknoglobals
...
 var itemDefinitionFormats = []fabricitem.DefinitionFormat{ //nolint:gochecknoglobals
 	{
-		Type:  fabricitem.DefinitionFormatDefault,
-		API:   "",
-		Paths: []string{"map.json"},
+		Type: fabricitem.DefinitionFormatDefault,
+		API:  "",
+		Paths: []string{
+			"map.json",
+			"queries/layerSource-*.kql",
```
OrBaubergMicrosoft, line 36: does this pattern (with layerSource-<layerSourceId>.kql) work? can you try even in a UT?

internal/services/notebook/base.go
```diff
@@ -28,13 +28,28 @@ var ItemTypeInfo = tftypeinfo.TFTypeInfo{ //nolint:gochecknoglobals
 
 var itemDefinitionFormats = []fabricitem.DefinitionFormat{ //nolint:gochecknoglobals
 	{
-		Type:  "ipynb",
-		API:   "ipynb",
-		Paths: []string{"notebook-content.ipynb"},
+		Type: "ipynb",
+		API:  "ipynb",
+		Paths: []string{
+			"*.ipynb",
```
OrBaubergMicrosoft, line 34: why * and not notebook content?

internal/services/notebook/resource_notebook_test.go
```diff
@@ -426,3 +432,57 @@ func TestAcc_NotebookDefinitionPYResource_CRUD(t *testing.T) {
 	},
 	))
 }
+
+func TestAcc_NotebookDefinitionRResource_CRUD(t *testing.T) {
```
OrBaubergMicrosoft: not sure if we need an acc test foreach definition format, since it is just for our validation, maybe keep them UTs?

internal/services/sparkjobdefinition/base.go
```diff
@@ -32,4 +32,13 @@ var itemDefinitionFormats = []fabricitem.DefinitionFormat{ //nolint:gochecknoglo
 		API:   "SparkJobDefinitionV1",
 		Paths: []string{"SparkJobDefinitionV1.json"},
 	},
+	{
+		Type: "SparkJobDefinitionV2",
+		API:  "SparkJobDefinitionV2",
+		Paths: []string{
+			"SparkJobDefinitionV1.json",
+			"Main/*",
```
OrBaubergMicrosoft: https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions/spark-job-definition#:~:text=uploading%20Python%20(.py)%20and%20R%20(.R)%20files%20only.
can we enforce this here?

internal/services/sqldatabase/base.go
```diff
@@ -35,6 +35,6 @@ var itemDefinitionFormats = []fabricitem.DefinitionFormat{ //nolint:gochecknoglo
 	{
 		Type:  "sqlproj",
 		API:   "sqlproj",
-		Paths: []string{"*.sqlproj", "*.sql"},
+		Paths: []string{"*.sqlproj", "*.sql", ".sharedqueries/*.sql"},
```
OrBaubergMicrosoft, line 38: according to docs sqldb.sqlproj
https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions/sql-database-definition#:~:text=Description-,sqldb.sqlproj,-SQL%20database%20project

internal/services/notebook/base.go
catalina-gav: because this is the example they have now in the api docs:

internal/services/digitaltwinbuilder/resource_digital_twin_builder.go
catalina-gav: If you refer to using PatternsIfAttributeIsOneOf here instead of OneOf/OneOfWithDescriptionIfAttributeIsOneOf, yes. Our definition path rules contain wildcard patterns (for example, EntityTypes/*), which are pattern rules, not exact filenames. Exact-match validators only accept literal strings, so a valid file like EntityTypes/139950578358348.json is incorrectly rejected. Pattern-based validation is required so * is evaluated as wildcard matching, which aligns provider behavior with the documented definition path rules.

internal/services/sqldatabase/base.go
catalina-gav: But this is how it's already been enforced in our provider to accept any name, wouldn't enforcing the sqldb name now be a breaking change?

internal/services/fabricmap/base.go
catalina-gav: added UT, it works

OrBaubergMicrosoft (approved):

PR #1012, opened 2026-06-16
fix(connection): credential rotation forcing replacement instead of a…

CatalinDabuleanu:
…n in-place update

# 📥 Pull Request

Fixed `fabric_connection` credential rotation forcing replacement instead of an in-place update on `VirtualNetworkGateway` connections


internal/services/connection/schema.go
```diff
@@ -153,6 +153,9 @@ func itemSchema(ctx context.Context, isList bool) superschema.Schema { //revive:
 				Resource: &schemaR.BoolAttribute{
 					Optional: true,
 					Computed: true,
+					PlanModifiers: []planmodifier.Bool{
+						boolplanmodifier.UseStateForUnknown(),
```
OrBaubergMicrosoft, line 157: did you try this fix locally?
are there any other similar cases across the provider?

OrBaubergMicrosoft (approved):

PR #1069, opened 2026-08-12
feat(item_job_scheduler): expose local_time_zone_id on schedule confi…

### LINKED ISSUE #1052: [FEAT] Expose local_time_zone_id on fabric_item_job_scheduler
### 🚀 Feature description

When I manage Fabric item job schedules with Terraform, I want to set configuration.local_time_zone_id so I can run schedules in any Windows timezone supported by the Fabric Job Scheduler API, instead of always using Central Standard Time.

Now the Fabric REST API accepts localTimeZoneId on all schedule config types (Cron, Daily, Weekly, Monthly), but fabric_item_job_scheduler hardcodes "Central Standard Time" on create/update and does not expose the field in schema or state. API responses that include localTimeZoneId are also dropped on read.

### 🔈 Motivation

Schedules for teams outside US Central cannot be expressed correctly. Daily/Weekly/Monthly times are interpreted in the schedule timezone, so a hard-coded CST causes jobs to run at the wrong local time. Workarounds require leaving Terraform or accepting incorrect run times.

### 🛰 Alternatives

- Leave timezone hardcoded (current behavior): not acceptable for non-CST workloads.
- Create schedules outside Terraform / via REST or portal: breaks IaC consistency.
- Encode timezone only via start/end UTC offsets: does not replace localTimeZoneId for recurrence/times interpretation per the API.

### 🚧 Potential Configuration / Desired Solution
```terraform
resource "fabric_item_job_scheduler" "example" {
  item_id      = "11111111-1111-1111-1111-111111111111"
  job_type     = "Execute"
  enabled      = true
  configuration = {
    start_date_time    = "2025-11-11T10:00:00Z"
    end_date_time      = "2025-11-12T10:00:00Z"
    local_time_zone_id = "W. Europe Standard Time" # Windows timezone ID
    type               = "Daily"
    times              = ["10:00"]
  }
}
```
### 📎 Additional context

API reference:
- https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler/create-item-schedule
- localTimeZoneId: Windows timezone IDs from
  https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/default-time-zones
  (e.g. "Central Standard Time", "UTC", "W. Europe Standard Time") — not IANA names like Europe/Amsterdam.

Provider code today:
- Hardcoded in internal/services/itemjobscheduler/models.go (requestCreateJobSchedule / requestUpdateJobSchedule):
  localTimeZoneID := "Central Standard Time"
- configurationModel and configurationSchema have no local_time_zone_id
- configurationModel.set() never maps LocalTimeZoneID from the API response
- Well-known test data in tools/scripts/Set-WellKnown.ps1 also uses "Central Standard Time"

Suggested implementation notes:
- Add configuration.local_time_zone_id to schema, model, create/update, and read.
- Prefer Optional + default "Central Standard Time" for backward compatibility with existing configs; Required would be a breaking change.
- Validate via docs link to Default Time Zones (or light OneOf); avoid maintaining a full TZ enum unless desired.
- Update examples, tests, and regenerate docs (do not hand-edit ./docs).

### ☑️ Acceptance Criteria

- [ ] configuration.local_time_zone_id is available on fabric_item_job_scheduler (resource + data sources that expose configuration)
- [ ] Create/Update send the configured value to the Fabric API (no hardcoded CST)
- [ ] Read/import populate local_time_zone_id from the API
- [ ] Existing configs without the attribute keep working (default CST) OR breaking change is documented
- [ ] Examples and generated docs show the attribute with Windows timezone ID guidance
- [ ] Unit/acceptance tests cover at least one non-CST timezone

### 🔰 Code of Conduct

OrBaubergMicrosoft:
## Description

Adds `configuration.local_time_zone_id` to the `fabric_item_job_scheduler` resource
and to both the singular and plural data sources.

Previously the provider hardcoded `"Central Standard Time"` as a local literal in
`requestCreateJobSchedule.set()` and `requestUpdateJobSchedule.set()` and sent it on
every create/update for all four schedule types (Cron, Daily, Weekly, Monthly). The
field is marked `REQUIRED` by the Fabric API, so something always had to be sent, but
there was no way for a user to influence it. On the read path
`configurationModel.set()` never mapped `LocalTimeZoneID`, so the value returned by the
API was silently dropped and never surfaced in state, refresh, or import — meaning
drift on that field was invisible.

Schedules for teams outside US Central therefore ran at the wrong local time, since
Daily/Weekly/Monthly `times` are interpreted in the schedule's timezone.

## Changes

- `models.go` — added `LocalTimeZoneID` to `configurationModel`, mapped it in
  `configurationModel.set()`, and both request builders now use the configured value.
- `schema.go` — new `configuration.local_time_zone_id` attribute:
  `Optional` + `Computed` + `Default` on the resource, `Computed` on the data sources.
- `base.go` — exported `DefaultLocalTimeZoneID` constant.
- `tools/scripts/Set-WellKnown.ps1` — exports `configurationLocalTimeZoneId` on the
  `ItemJobScheduler` well-known entry so acceptance tests can assert the value.
- Examples + regenerated docs.

## Backward compatibility

Non-breaking. The attribute defaults to `Central Standard Time`, which is exactly the
value the provider sent before, so existing configurations produce an empty plan. Note
that the value must be a **Windows** timezone identifier (e.g. `W. Europe Standard Time`),
not an IANA name like `Europe/Amsterdam`.

## Testing

- Unit: CRUD asserts the default is applied when omitted, and that a non-CST value
  (`W. Europe Standard Time`) round-trips on update; data source tests assert the value
  is read back rather than dropped.
- Acceptance: resource CRUD covers both default and non-CST; both data source tests
  assert against the well-known fixture value.
- `task docs`, `golangci-lint`, unit (6/6) and acceptance (3/3) all pass.

Closes #1052


.changes/unreleased/added-20260812-000000.yaml
```diff
@@ -0,0 +1,6 @@
+kind: added
+body: Added `configuration.local_time_zone_id` to `fabric_item_job_scheduler` resource
+  and data sources. Defaults to `Central Standard Time` when not specified.
```
badeamarjieh: remove, irrelevant for changelog

badeamarjieh (approved):

badeamarjieh (approved):

Issue #903, opened 2026-04-01
[FEAT] Support Workspace IP Firewall Rules


aslatter:
### 🚀 Feature description

>>>
Workspace IP firewall rules let workspace admins control access to their Microsoft Fabric workspace by allowing connections only from trusted public IP addresses. By setting up a simple allow list, you can prevent unauthorized traffic from reaching your workspace. This feature adds an extra layer of protection in addition to identity and role-based access controls.
>>>

From: https://learn.microsoft.com/en-us/fabric/security/security-workspace-level-firewall-overview

### 🚧 Potential Configuration / Desired Solution

```tf
resource "fabric_workspace_ip_firewall_rules" "example" {
  workspace_id = fabric_workspace.example.id

  rule {
    name = "a rule"
    value = "192.168.1.0/24"
  }

  rule {
    name = "another rule"
    value = "10.0.0.0/8"
  }
}
```

### 📎 Additional context

API documentation: 
https://learn.microsoft.com/en-us/fabric/security/security-workspace-level-firewall-set-up?tabs=api-1#configure-workspace-ip-firewall-rules

### 🔰 Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

Halamish51:
Hi @aslatter, 

Thanks for your feedback. We are reviewing it and will get back to you.

OrBaubergMicrosoft:
https://learn.microsoft.com/en-us/rest/api/fabric/core/workspaces/get-firewall-rules?tabs=HTTP
https://learn.microsoft.com/en-us/rest/api/fabric/core/workspaces/set-firewall-rules?tabs=HTTP

PR #1081, opened 2026-08-16
feat(workspace_firewall_rules): onboard resource and data source

### LINKED ISSUE #1079: [RS] fabric_workspace_firewall_rules
### 📝 Description

When managing Microsoft Fabric infrastructure as code, I want to declare the set of IP firewall rules that guard a workspace's inbound access, so I can enforce a trusted-IP allow list consistently across environments without clicking through the portal.

Workspace IP firewall rules let workspace admins restrict access to a Fabric workspace to a set of trusted public IP addresses, layered on top of identity and role-based access controls.

> [!IMPORTANT]
> Firewall rules are **only enforced** when the workspace's network communication policy has `inbound.public_access_rules.default_action` set to `Deny`. Otherwise the rules are stored but have no effect.

Paired data source issue: #1080
Originally raised in #903.

### 🔬 Details / References

- Resource Name: `fabric_workspace_firewall_rules`
- Suggested package: `internal/services/workspacefwr`
- REST endpoint: `GET|PUT /v1/workspaces/{workspaceId}/networking/communicationPolicy/inbound/firewall`
- API documentation:
  - [Workspaces - Get Firewall Rules](https://learn.microsoft.com/rest/api/fabric/core/workspaces/get-firewall-rules)
  - [Workspaces - Set Firewall Rules](https://learn.microsoft.com/rest/api/fabric/core/workspaces/set-firewall-rules)
  - [Workspace-level firewall overview](https://learn.microsoft.com/fabric/security/security-workspace-level-firewall-overview)
  - [Set up workspace-level firewall](https://learn.microsoft.com/fabric/security/security-workspace-level-firewall-set-up)
- Estimated complexity/effort: **moderate**
- Resource category: **non-item resource**
- Implementation pattern: **Pattern A** — workspace policy singleton (`fabcore.WorkspacesClient` sub-endpoint, no entity ID, delete = reset)
- Canonical reference: [`internal/services/workspacencp/`](https://github.com/microsoft/terraform-provider-fabric/tree/main/internal/services/workspacencp)
- Preview: ✅ The API is explicitly marked **Preview** in the SDK — requires `IsPreview: true` in `base.go` and `fabricitem.IsPreviewMode` gating
- SPN support: ✅ User, Service Principal, and Managed Identity
- Related resources/data-sources:
  - Paired data source: #1080
  - `fabric_workspace_network_communication_policy` (`internal/services/workspacencp`) — **enforcement precondition**
  - `fabric_workspace` (`internal/services/workspace`)
  - `fabric_workspace_managed_private_endpoint` (`internal/services/workspacempe`)

#### ⛔ Blocking dependency — SDK bump required

`go.mod` currently pins `github.com/microsoft/fabric-sdk-go v0.19.0`, which **does not** contain the firewall APIs. They were introduced in **v0.20.0**.

| SDK version | `GetFirewallRules` / `SetFirewallRules` |
| ----------- | --------------------------------------- |
| v0.19.0 (current) | ❌ Not present |
| v0.20.0 | ✅ `fabric/core` `WorkspacesClient` |

➡️ Run `task deps:up` (or bump explicitly) to `fabric-sdk-go >= v0.20.0` **before** starting implementation.

#### Tenant prerequisites

- **Tenant setting** `Configure workspace IP firewall rules` must be enabled (enabled by default; a Fabric admin can disable it).
- The workspace must be assigned to a **Fabric capacity or Trial capacity**.
- First-time workspace-level networking setup in a tenant may require re-registering the `Microsoft.Fabric` resource provider in Azure.

#### SDK contract

```go
// github.com/microsoft/fabric-sdk-go/fabric/core  (alias: fabcore)
// client: fabcore.NewClientFactoryWithClient(*pConfigData.FabricClient).NewWorkspacesClient()

func (client *WorkspacesClient) GetFirewallRules(
    ctx context.Context,
    workspaceID string,
    options *WorkspacesClientGetFirewallRulesOptions,
) (WorkspacesClientGetFirewallRulesResponse, error)

func (client *WorkspacesClient) SetFirewallRules(
    ctx context.Context,
    workspaceID string,
    firewallRulesRequest InboundFirewallConfiguration,
    options *WorkspacesClientSetFirewallRulesOptions,
) (WorkspacesClientSetFirewallRulesResponse, error)
```

DTOs:

```go
// InboundFirewallConfiguration - the complete set of firewall rules for a workspace's inbound access protection.
type InboundFirewallConfiguration struct {
    // A maximum of 256 rules can be specified per workspace.
    Rules []FirewallRule
}

// FirewallRule - a rule identified by its name, containing a single IP, an IP range, or a CIDR IP as its value.
type FirewallRule struct {
    // REQUIRED; up to 128 characters, unique per workspace.
    DisplayName *string
    // REQUIRED; a single IP address, an IP range in start-end format, or CIDR notation.
    Value *string
}
```

- `WorkspacesClientGetFirewallRulesResponse` embeds `InboundFirewallConfiguration`
- `WorkspacesClientSetFirewallRulesResponse` is **empty** (no body)
- Fake server support exists: `fabcore/fake` → `WorkspacesServer.GetFirewallRules` / `WorkspacesServer.SetFirewallRules`

#### Permissions

| Operation | Workspace role | Delegated scope |
| --------- | -------------- | --------------- |
| `GetFirewallRules` | Viewer or higher | `Workspace.Read.All` or `Workspace.ReadWrite.All` |
| `SetFirewallRules` | **Admin** | `Workspace.ReadWrite.All` |

#### Service-side constraints

| Constraint | Detail |
| ---------- | ------ |
| Max rules | 256 per workspace |
| `displayName` | 1–128 characters, **must be unique** per workspace (duplicates rejected) |
| `value` formats | Single IP, IP range (`start-end`), or CIDR |
| **RFC 1918 excluded** | Private ranges are **not supported** — `10.0.0.0/8`, `172.16.0.0/12`–`172.31.0.0/16`, `192.168.0.0/16` are rejected |
| Whitespace | Spaces are not allowed inside IP values |
| Private endpoint VMs | Public IPs of VMs on VNets with private endpoints cannot be used |

### 🎨 Terraform Design Notes

**Why a standalone resource rather than folding into `fabric_workspace_network_communication_policy`:**

The *admin/read* SDK model nests firewall inside inbound (`fabadmin.NetworkCommunicationPolicyInboundDetails.Firewall`), but the *core/write* model does **not** — `fabcore.InboundRules` has only `PublicAccessRules`. Firewall rules are a separate endpoint pair with their own permission levels and their own **Preview** status, while `workspacencp` is GA (`IsPreview: false`). Folding them together would require either regressing a GA resource behind `FABRIC_PREVIEW` or inventing per-attribute preview gating (no precedent — `fabricitem.IsPreviewMode` gates at resource level). It would also introduce partial-apply risk (2× PUT per apply) and silently expand `workspacencp`'s destructive `Delete`-resets-everything behavior to wipe up to 256 firewall rules.

> If `fabcore.InboundRules` ever gains a `Firewall` field (single combined PUT), revisit and consider consolidating via deprecation.

**Design decisions:**

1. **Whole-set resource, not per-rule.** `SetFirewallRules` is a **PUT that fully replaces** all rules. Per-rule resources are not viable — the entire rule set must be a single `rules` list attribute.
2. **Read-after-write.** `SetFirewallRules` returns no body → follow every Create/Update with `GetFirewallRules` to populate state (same as [`workspacencp`](https://github.com/microsoft/terraform-provider-fabric/blob/main/internal/services/workspacencp/resource_workspace_network_communication_policy.go)).
3. **No DELETE endpoint.** Implement `Delete` as `SetFirewallRules(ctx, workspaceID, fabcore.InboundFirewallConfiguration{}, nil)` and emit a reset warning, mirroring `workspacencp`'s `Delete`.
4. **`workspace_id` is immutable** → `stringplanmodifier.RequiresReplace()`, `customtypes.UUIDType{}`.
5. **Enforcement-precondition warning.** In `Create`/`Update`, call `GetNetworkCommunicationPolicy` and if `inbound.public_access_rules.default_action != Deny`, emit a **warning** diagnostic that the rules are stored but not enforced. Warning, not error — users may legitimately stage rules before flipping to `Deny`.
6. **Validators:**
   - `rules` — `listvalidator.SizeAtMost(256)`
   - `rules[*].display_name` — `stringvalidator.LengthBetween(1, 128)`
   - `rules[*].value` — validate single IP / `start-end` range / CIDR, **and reject RFC 1918 private ranges** with an actionable message (the service rejects them, so failing at plan time is far better UX)
   - Consider a set-level validator rejecting duplicate `display_name` values at plan time
7. **HCL naming** — SDK `DisplayName` → `display_name` (repo convention), not `name`.
8. **Import** — composite identity is just `workspace_id`. Note that `workspacencp` does not currently implement `ImportState`; decide whether to align or add import here.

### 🚧 Potential Terraform Configuration / Desired Solution

> [!NOTE]
> Only **public** IP addresses are accepted. The example below uses RFC 5737 documentation ranges. RFC 1918 private ranges (`10.x`, `172.16–31.x`, `192.168.x`) are rejected by the service.

```terraform
resource "fabric_workspace_network_communication_policy" "example" {
  workspace_id = fabric_workspace.example.id

  inbound = {
    public_access_rules = {
      default_action = "Deny"
    }
  }
}

resource "fabric_workspace_firewall_rules" "example" {
  workspace_id = fabric_workspace.example.id

  rules = [
    {
      display_name = "corp-egress"
      value        = "203.0.113.0/24"
    },
    {
      display_name = "vpn-range"
      value        = "198.51.100.10-198.51.100.20"
    },
    {
      display_name = "single-host"
      value        = "192.0.2.42"
    },
  ]

  # Rules are only enforced once inbound public access is set to Deny.
  depends_on = [fabric_workspace_network_communication_policy.example]
}
```

### 📎 Additional context

- Supersedes the design intent of #903 (which was filed as a `[FEAT]` but is in fact a new resource + data source).
- Recommend implementing together with #1080 in a single PR since they share `schema.go`, `models.go`, and `base.go`.
- Cross-link the `MarkdownDescription` of both this resource and `fabric_workspace_network_communication_policy` so the enforcement precondition is discoverable from either side.

#### Well-known / acceptance test infrastructure

- **`internal/testhelp/wellknown.go` requires no changes** — it is a generic `map[string]any` accessor.
- **Resource acceptance test:** no new well-known entry needed. Follow the `workspacencp` pattern — use `testhelp.WellKnown()["Capacity"]` plus `testhelp.TestAccWorkspaceResource(t, capacityID)` to spin up an ephemeral workspace per test run.
- **Data source acceptance test:** needs a workspace with at least one pre-seeded firewall rule. Add a `Set-FabricWorkspaceFirewallRules` helper to [`tools/scripts/Set-WellKnown.ps1`](https://github.com/microsoft/terraform-provider-fabric/blob/main/tools/scripts/Set-WellKnown.ps1) (modelled on the existing `Set-OutboundNetworkPolicy`) and seed rules onto `WorkspaceDS`, exposing them as `$wellKnown['WorkspaceDS'].firewallRules`.
- 🚨 **Do NOT set `inbound.publicAccessRules.defaultAction = Deny` on any shared well-known workspace.** Doing so activates enforcement and will lock the CI runner out of that workspace unless its egress IP is allow-listed — and recovery is API-only. Leaving the default action at `Allow` means the seeded rules are stored but unenforced, which is exactly what the read tests need and carries zero lockout risk.
- Run `@wellknown-setup` before implementation.

### ☑️ Acceptance Criteria

- [ ] `fabric-sdk-go` bumped to `>= v0.20.0` and the module builds
- [ ] `fabric_workspace_firewall_rules` resource registered in `internal/provider/provider.go`
- [ ] Resource is gated behind the provider `preview` flag (`IsPreview: true`)
- [ ] `Create` and `Update` both issue `SetFirewallRules` followed by a `GetFirewallRules` read-after-write
- [ ] `Delete` clears all rules by PUTting an empty `InboundFirewallConfiguration` and surfaces a reset warning
- [ ] `workspace_id` changes force resource replacement
- [ ] A warning diagnostic is raised when `inbound.public_access_rules.default_action != "Deny"`
- [ ] `rules` is capped at 256 entries; `display_name` is validated to 1–128 characters
- [ ] `value` accepts a single IP, an IP range (`start-end`), and CIDR notation
- [ ] `value` rejects RFC 1918 private ranges at plan time with an actionable error message
- [ ] Duplicate `display_name` values are rejected at plan time
- [ ] Removing the resource from config resets the workspace to zero firewall rules without altering `default_action`
- [ ] Unit tests cover happy path and error paths (unauthorized, workspace not found, invalid rule value, private IP range, duplicate name, >256 rules) using `fabcore/fake` `WorkspacesServer` handlers
- [ ] Acceptance tests cover create, update (add/remove/reorder rules), and destroy
- [ ] `examples/resources/fabric_workspace_firewall_rules/` contains `resource.tf` and `import.sh` (if import is implemented)
- [ ] `task docs` regenerated and `task lint` passes

### ✅ Definition of Done

- [ ] Data Transfer Objects (DTOs)
- [ ] Resource Implementation
- [ ] Resource Added to Provider
- [ ] Unit Tests for Happy path
- [ ] Unit Tests for Error path
- [ ] Acceptance Tests
- [ ] Example in the ./examples folder
- [ ] Schema documentation in code
- [ ] Updated auto-generated provider docs with `task docs`

### 🔰 Code of Conduct

### LINKED ISSUE #1080: [DS] fabric_workspace_firewall_rules
### 📝 Description

When auditing or composing Microsoft Fabric infrastructure as code, I want to read the IP firewall rules currently configured on a workspace, so I can reference the existing allow list in other resources, validate compliance, or inspect a workspace that Terraform does not own.

Related resource issue: #1079
Originally raised in #903.

### 🔬 Details / References

- Data Source Name: `fabric_workspace_firewall_rules`
- Suggested package: `internal/services/workspacefwr` (same package as the resource)
- REST endpoint: `GET /v1/workspaces/{workspaceId}/networking/communicationPolicy/inbound/firewall`
- API documentation:
  - [Workspaces - Get Firewall Rules](https://learn.microsoft.com/rest/api/fabric/core/workspaces/get-firewall-rules)
  - [Workspace-level firewall overview](https://learn.microsoft.com/fabric/security/security-workspace-level-firewall-overview)
- Estimated complexity/effort: **easy**
- Resource category: **non-item resource**
- Implementation pattern: **Pattern A** — workspace policy singleton (`fabcore.WorkspacesClient` sub-endpoint, no entity ID)
- Canonical reference: [`internal/services/workspacencp/data_workspace_network_communication_policy.go`](https://github.com/microsoft/terraform-provider-fabric/blob/main/internal/services/workspacencp/data_workspace_network_communication_policy.go)
- Preview: ✅ The API is explicitly marked **Preview** in the SDK — requires `IsPreview: true` in `base.go` and `fabricitem.IsPreviewMode` gating
- SPN support: ✅ User, Service Principal, and Managed Identity
- Related resources/data-sources:
  - Paired resource: #1079
  - `fabric_workspace_network_communication_policy` (`internal/services/workspacencp`) — determines whether the rules are actually enforced
  - `fabric_workspace` (`internal/services/workspace`)

#### ⛔ Blocking dependency — SDK bump required

`go.mod` currently pins `github.com/microsoft/fabric-sdk-go v0.19.0`, which **does not** contain the firewall APIs. They were introduced in **v0.20.0**. Run `task deps:up` (or bump explicitly) to `fabric-sdk-go >= v0.20.0` before starting implementation.

#### Tenant prerequisites

- **Tenant setting** `Configure workspace IP firewall rules` must be enabled (enabled by default; a Fabric admin can disable it).
- The workspace must be assigned to a **Fabric capacity or Trial capacity**.

#### SDK contract

```go
// github.com/microsoft/fabric-sdk-go/fabric/core  (alias: fabcore)
// client: fabcore.NewClientFactoryWithClient(*pConfigData.FabricClient).NewWorkspacesClient()

func (client *WorkspacesClient) GetFirewallRules(
    ctx context.Context,
    workspaceID string,
    options *WorkspacesClientGetFirewallRulesOptions,
) (WorkspacesClientGetFirewallRulesResponse, error)
```

DTOs:

```go
// InboundFirewallConfiguration - the complete set of firewall rules for a workspace's inbound access protection.
type InboundFirewallConfiguration struct {
    // A maximum of 256 rules can be specified per workspace.
    Rules []FirewallRule
}

// FirewallRule - a rule identified by its name, containing a single IP, an IP range, or a CIDR IP as its value.
type FirewallRule struct {
    // REQUIRED; up to 128 characters, unique per workspace.
    DisplayName *string
    // REQUIRED; a single IP address, an IP range in start-end format, or CIDR notation.
    Value *string
}
```

- `WorkspacesClientGetFirewallRulesResponse` embeds `InboundFirewallConfiguration`
- Fake server support exists: `fabcore/fake` → `WorkspacesServer.GetFirewallRules`

#### Permissions

| Operation | Workspace role | Delegated scope |
| --------- | -------------- | --------------- |
| `GetFirewallRules` | Viewer or higher | `Workspace.Read.All` or `Workspace.ReadWrite.All` |

### 🎨 Terraform Design Notes

1. **Single-workspace read only.** The API is scoped to one workspace and there is no list/pager across workspaces, so there is **no plural (list) data source** to build. A tenant-wide view exists only via the `fabadmin` network communication policy listing (`NetworkCommunicationPolicyInboundDetails.Firewall`) — out of scope for this issue.
2. **Inputs:** `workspace_id` (Required, `customtypes.UUIDType{}`).
3. **Outputs:** `rules` — a computed list of objects with `display_name` and `value`. All attributes `Computed: true`.
4. **Shared schema.** Use `superschema` with `Common` / `DataSource` blocks so the resource (#1079) and this data source share one schema definition, exactly as [`internal/services/workspacencp/schema.go`](https://github.com/microsoft/terraform-provider-fabric/blob/main/internal/services/workspacencp/schema.go) does.
5. **Not-found handling:** map `fabcore.ErrCommon.EntityNotFound` via `utils.IsErrNotFound` / `utils.GetDiagsFromError`.
6. **HCL naming** — SDK `DisplayName` → `display_name` (repo convention), not `name`.
7. **Documentation note:** the data source should state that returned rules are only enforced when the workspace's `inbound.public_access_rules.default_action` is `Deny`, and link to `fabric_workspace_network_communication_policy`.

### 🚧 Potential Terraform Configuration / Desired Solution

```terraform
data "fabric_workspace_firewall_rules" "example" {
}

output "firewall_rule_values" {
  value = [for rule in data.fabric_workspace_firewall_rules.example.rules : rule.value]
}
```

### 📎 Additional context

- Supersedes the design intent of #903 (which was filed as a `[FEAT]` but is in fact a new resource + data source).
- Recommend implementing together with #1079 in a single PR since they share `schema.go`, `models.go`, and `base.go`.

#### Well-known / acceptance test infrastructure

- **`internal/testhelp/wellknown.go` requires no changes** — it is a generic `map[string]any` accessor.
- The acceptance test needs a workspace with **at least one pre-seeded firewall rule** so the assertions are meaningful (a zero-rule workspace only proves the call succeeds).
- Add a `Set-FabricWorkspaceFirewallRules` helper to [`tools/scripts/Set-WellKnown.ps1`](https://github.com/microsoft/terraform-provider-fabric/blob/main/tools/scripts/Set-WellKnown.ps1) (modelled on the existing `Set-OutboundNetworkPolicy`, PUT to `workspaces/{id}/networking/communicationPolicy/inbound/firewall`) and seed rules onto `WorkspaceDS`, exposing them as `$wellKnown['WorkspaceDS'].firewallRules` for the test to assert against.
- Seed values must be **public** IP ranges — RFC 1918 private ranges (`10.x`, `172.16–31.x`, `192.168.x`) are rejected by the service. Use RFC 5737 documentation ranges such as `203.0.113.0/24`.
- 🚨 **Do NOT set `inbound.publicAccessRules.defaultAction = Deny` on `WorkspaceDS` or any other shared well-known workspace.** That activates enforcement and would lock the CI runner out of the workspace unless its egress IP is allow-listed — recovery is API-only. Leaving the default action at `Allow` means the seeded rules are stored but unenforced, which is exactly what this read test needs and carries zero lockout risk.
- Run `@wellknown-setup` before implementation.

### ☑️ Acceptance Criteria

- [ ] `fabric-sdk-go` bumped to `>= v0.20.0` and the module builds
- [ ] `fabric_workspace_firewall_rules` data source registered in `internal/provider/provider.go`
- [ ] Data source is gated behind the provider `preview` flag (`IsPreview: true`)
- [ ] `workspace_id` is `Required` and validated as a UUID
- [ ] `rules` is returned as a computed list of objects with `display_name` and `value`
- [ ] A workspace with zero firewall rules returns an empty list rather than an error
- [ ] A non-existent or inaccessible workspace produces a clear not-found / unauthorized diagnostic
- [ ] Schema is shared with the resource (#1079) via a single `superschema` definition
- [ ] Unit tests cover happy path and error paths (not found, unauthorized) using `fabcore/fake` `WorkspacesServer` handlers
- [ ] Acceptance test reads the well-known workspace and asserts the seeded rule values
- [ ] `examples/data-sources/fabric_workspace_firewall_rules/data-source.tf` added
- [ ] `task docs` regenerated and `task lint` passes

### ✅ Definition of Done

- [ ] Data Transfer Objects (DTOs)
- [ ] Data-Source Implementation
- [ ] Data-Source Added to Provider
- [ ] Unit Tests for Happy path
- [ ] Unit Tests for Error path
- [ ] Acceptance Tests
- [ ] Example in the ./examples folder
- [ ] Schema documentation in code
- [ ] Updated auto-generated provider docs with `task docs`

### 🔰 Code of Conduct

OrBaubergMicrosoft:
### ❓ What are you trying to address

Closes #1079
Closes #1080
Supersedes the design intent of #903

Microsoft Fabric supports workspace-level IP firewall rules, letting workspace admins restrict inbound access to an allow list of trusted public IP addresses. The provider had no way to manage them — the only options were the Fabric portal or raw REST calls.

### ✨ Description of new changes

Adds `fabric_workspace_firewall_rules` as both a **resource** and a **data source**, in a new `workspacefwr` package.

**Implementation**

- Non-item resource, **Pattern A** (workspace policy singleton), modeled on `workspacencp` and `workspaceogr`
- Backed by `fabcore.WorkspacesClient.GetFirewallRules` / `SetFirewallRules` (`GET|PUT /v1/workspaces/{workspaceId}/networking/communicationPolicy/inbound/firewall`)
- Gated behind provider **preview** mode (`IsPreview: true`) — the underlying API is a Preview release
- SPN supported

**Behavior driven by the API's semantics**

| API constraint | Provider behavior |
| --- | --- |
| `PUT` fully replaces the rule set | `rules` is a single whole-set attribute; per-rule resources are not viable |
| `Set` returns an empty body | Read-after-write: every Create/Update is followed by a `GET` |
| No `DELETE` endpoint | `Delete` PUTs an empty configuration and emits a reset warning |
| No ordering guarantee on read | `rules` is a **Set**, not a List, so reordered responses don't produce diffs |
| Max 256 rules / name ≤128 chars, unique | `setvalidator.SizeAtMost(256)`, `stringvalidator.LengthBetween(1, 128)` |

`workspace_id` is immutable (`RequiresReplace`).

> [!NOTE]
> Firewall rules are only *enforced* when `inbound.public_access_rules.default_action` on `fabric_workspace_network_communication_policy` is `Deny`. This is documented in the schema description and wired into the example via `depends_on`.

**Why a separate resource rather than folding into `fabric_workspace_network_communication_policy`:** the `fabcore` *write* model (`InboundRules`) has no `Firewall` field — these are separate endpoints with different permission levels (Get: viewer+, Set: admin) and different preview status (`workspacencp` is GA). Merging would have meant either regressing a GA resource behind `FABRIC_PREVIEW` or inventing per-attribute preview gating, plus partial-apply risk from two PUTs per apply.

**Tests** — 8 unit tests, 79.1% coverage

- `_Attributes` — missing/unexpected args, invalid UUID, incomplete rule, `display_name` >128
- `_CRUD` — create → update → remove all rules
- `_CreateError` / `_UpdateError` / `_DeleteError` — API failures on each write operation via `fabfake.SetResponseError`
- `_RuleOrderIsIgnored` — API returns rules reversed; asserts state matches config and the refresh plan is empty (`plancheck.ExpectEmptyPlan()`)
- `DataSource` — read + config validation errors; `DataSource_Error` — API failure


internal/services/workspacefwr/schema.go
```diff
@@ -0,0 +1,116 @@
...
+					MarkdownDescription: "A set of rules that define the IP addresses permitted for inbound access to the Workspace. " +
+						"Rules are only enforced when `inbound.public_access_rules.default_action` of the `fabric_workspace_network_communication_policy` is set to `Deny`. " +
+						"Only public IP addresses are supported.",
+					Computed: true,
+				},
+				Resource: &schemaR.SetNestedAttribute{
+					Optional: true,
+					Validators: []validator.Set{
+						setvalidator.SizeAtMost(maxRules),
+					},
```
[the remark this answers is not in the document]
OrBaubergMicrosoft: requires new validator, will skip this validation for now

internal/services/workspacefwr/resource_workspace_firewall_rules.go
```diff
@@ -0,0 +1,255 @@
...
+	var reqCreate requestSetWorkspaceFirewallRules
+
+	if resp.Diagnostics.Append(reqCreate.set(ctx, plan)...); resp.Diagnostics.HasError() {
+		return
+	}
+
+	_, err := r.client.SetFirewallRules(ctx, plan.WorkspaceID.ValueString(), reqCreate.InboundFirewallConfiguration, nil)
+	if resp.Diagnostics.Append(utils.GetDiagsFromError(ctx, err, utils.OperationCreate, nil)...); resp.Diagnostics.HasError() {
+		return
+	}
```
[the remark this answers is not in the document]
OrBaubergMicrosoft: aligning with repo's behavior in other communication policy APIs, leaving as is for now

internal/services/workspacefwr/data_workspace_firewall_rules_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/microsoft/terraform-provider-fabric/internal/common"
+	"github.com/microsoft/terraform-provider-fabric/internal/framework/customtypes"
+	"github.com/microsoft/terraform-provider-fabric/internal/testhelp"
+	"github.com/microsoft/terraform-provider-fabric/internal/testhelp/fakes"
+)
+
+var testDataSourceItemFQN, testDataSourceItemHeader = testhelp.TFDataSource(common.ProviderTypeName, itemTypeInfo.Type, "test")
+
+func TestUnit_WorkspaceFirewallRulesDataSource(t *testing.T) {
```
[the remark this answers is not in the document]
OrBaubergMicrosoft: We do not have a private links acc tests setup

.changes/unreleased/added-20260816-122000.yaml
```diff
@@ -0,0 +1,5 @@
+kind: added
+body: New Resource `fabric_workspace_firewall_rules` and Data Source `fabric_workspace_firewall_rules` to manage workspace inbound IP firewall rules.
```
OrBaubergMicrosoft: might need to split to two change logs

internal/services/workspacefwr/resource_workspace_firewall_rules.go
```diff
@@ -0,0 +1,255 @@
...
+	// Reset to defaults by sending an empty request body.
+	resp.Diagnostics.AddWarning(
+		"Resetting Firewall Rules to defaults",
+		fmt.Sprintf(
+			"Resource %s does not support deletion. All Firewall Rules will be removed from the Workspace.",
+			r.TypeInfo.Name,
+		),
+	)
+
+	reqDelete := fabcore.InboundFirewallConfiguration{}
```
OrBaubergMicrosoft: need to pass empty config instead of null

internal/services/workspacefwr/schema.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+			"rules": superschema.SuperSetNestedAttributeOf[firewallRuleModel]{
+				Common: &schemaR.SetNestedAttribute{
+					MarkdownDescription: "A set of rules that define the IP addresses permitted for inbound access to the Workspace. " +
+						"Rules are only enforced when `inbound.public_access_rules.default_action` of the `fabric_workspace_network_communication_policy` is set to `Deny`. " +
+						"Only public IP addresses are supported.",
+					Computed: true,
+				},
+				Resource: &schemaR.SetNestedAttribute{
+					Optional: true,
```
OrBaubergMicrosoft: should turn to required and remove default

.changes/unreleased/added-20260816-122000.yaml
```diff
@@ -0,0 +1,5 @@
+kind: added
+body: 'New Resource: `fabric_workspace_firewall_rules` to manage inbound IP firewall rules of a Workspace.'
```
badeamarjieh: Added `fabric_workspace_firewall_rules` resource

.changes/unreleased/added-20260816-122100.yaml
```diff
@@ -0,0 +1,5 @@
+kind: added
+body: 'New Data-Source: `fabric_workspace_firewall_rules` to read inbound IP firewall rules of a Workspace.'
```
badeamarjieh: Added `fabric_workspace_firewall_rules` data source

internal/services/workspacefwr/base.go
```diff
@@ -0,0 +1,14 @@
...
+
+package workspacefwr
+
+import "github.com/microsoft/terraform-provider-fabric/internal/pkg/tftypeinfo"
+
+var ItemTypeInfo = tftypeinfo.TFTypeInfo{ //nolint:gochecknoglobals
+	Name:           "Workspace Firewall Rules",
+	Type:           "workspace_firewall_rules",
+	DocsURL:        "https://learn.microsoft.com/fabric/security/security-workspace-level-firewall-overview",
+	IsPreview:      true,
```
badeamarjieh: false? the API is GA

internal/services/workspacefwr/models.go
```diff
@@ -0,0 +1,94 @@
...
+func (to *baseWorkspaceFirewallRulesModel) set(ctx context.Context, workspaceID string, from fabcore.InboundFirewallConfiguration) diag.Diagnostics {
+	to.WorkspaceID = customtypes.NewUUIDValue(workspaceID)
+
+	slice := make([]*firewallRuleModel, 0, len(from.Rules))
+
+	for _, prop := range from.Rules {
+		firewallRule := &firewallRuleModel{}
+		firewallRule.set(prop)
+		slice = append(slice, firewallRule)
+	}
```
badeamarjieh: we use a different pattern:

```
	for _, entity := range from.Rules {
		var entityModel firewallRuleModel
		entityModel.set(entity)
		slice = append(slice, &entityModel)
    }
```

internal/services/workspacefwr/models.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if diags := to.Rules.Set(ctx, slice); diags.HasError() {
+		return diags
+	}
+
+	return nil
```
badeamarjieh: return to.Rules.Set(ctx, slice) ?

badeamarjieh (changes requested):

internal/services/workspacefwr/schema.go
```diff
@@ -53,22 +50,15 @@ func itemSchema() superschema.Schema {
 					MarkdownDescription: "A set of rules that define the IP addresses permitted for inbound access to the Workspace. " +
 						"Rules are only enforced when `inbound.public_access_rules.default_action` of the `fabric_workspace_network_communication_policy` is set to `Deny`. " +
 						"Only public IP addresses are supported.",
-					Computed: true,
 				},
 				Resource: &schemaR.SetNestedAttribute{
-					Optional: true,
+					Required: true,
```
badeamarjieh, line 55: is it required?

badeamarjieh (changes requested):

internal/services/workspacefwr/schema.go
OrBaubergMicrosoft: The API accepts empty FW rules list
I think that no firewall rules should be handled explicitly via not having this resource at all (or TF destroy, if you once had it), instead of providing an empty list

badeamarjieh (approved):

PR #1106, opened 2026-09-10
fix(onelakedataaccesssecurity): preserve object_type for microsoft_entra_members after apply

### LINKED ISSUE #1044: [BUG] fabric_onelake_data_access_security with microsoft_entra_members fails after creation because of incosistent result after apply
### 🐛 What happened?

When using fabric_onelake_data_access_security with member type "microsoft_entra_members" instead of "fabric_item_members", you get an inconsistent result after apply because it cannot find the created element.

I suspect that the issue originates from the following situation: 

When creating, you provide object_id, object_type, tenant_id via the [Fabric API - Update-Create](https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-data-access-security/create-or-update-single-data-access-role?tabs=HTTP), but after apply, it checks the actual configuration via the [Fabric API - List](https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-data-access-security/list-data-access-roles?tabs=HTTP), which only returns object_id and tenant_id, without object_type. Thus it searches for an object with these 3 attributes in the returned list, which it cannot find because it list only contains objects with 2 attributes.

### 🔬 How to reproduce?

1. Create 2 Workspaces, 1 where the Lake House lives in (source), 1 where you want to access from (target)
2. Enable Workspace Identity in the target workspace
3. Create the resource with 

```
resource "fabric_onelake_data_access_security" "example" {
  item_id      = "11111111-1111-1111-1111-111111111111"
  role_name    = "example"
  decision_rules = [
    {
      effect = "Permit"
      permission = [
        {
          attribute_name              = "Path"
          attribute_value_included_in = ["*"]
        },
        {
          attribute_name              = "Action"
          attribute_value_included_in = ["Read"]
        }
      ]
    }
  ]
  members = {
    fabric_item_members = []
	microsoft_entra_members = [
      {
        object_id   = data.fabric_workspace.target_workspace.identity.service_principal_id
        object_type = "ManagedIdentity"
        tenant_id   = "33333333-3333-3333-3333-333333333333"
      }
    ]
  }
}
```

4. Plan & Apply
5. Apply fails after creating the resource, i.e. Role including members exist on Lakehouse, but Terraform State does not reflect it.

--
Infos:
* I tried different Object Types (ManagedIdentity, ServicePrincipal, User), all resulting in the same error
* fabric_item_members = [] is needed, otherwise the state hase "null" but the api returns "[]"

### 🏗️ Code Sample / Log

Error Message:
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ fabric_onelake_data_access_security.example,
│ provider "provider[\"registry.terraform.io/microsoft/fabric\"]" produced an
│ unexpected new value: .members.microsoft_entra_members: planned set element
│ cty.ObjectVal(map[string]cty.Value{"object_id":cty.StringVal("XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"),
│ "object_type":cty.StringVal("ManagedIdentity"),
│ "tenant_id":cty.StringVal("XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX")}) does
│ not correlate with any element in actual.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.

### 📷 Screenshots

_No response_

### 📈 Expected behavior

The check after creating the resource omits the "object type" and thus finds the entry in the list, resulting in a successfully registered apply.

1.12.0

1.15.8

### 🌌 Environment (OS)

Linux

### 📎 Additional context

_No response_

### 🔰 Code of Conduct

KRJNK:
# 📥 Pull Request

Fixes #1044.

`fabric_onelake_data_access_security` fails with `Provider produced inconsistent result after apply` whenever a role's `members` includes `microsoft_entra_members`. The Fabric API accepts `object_id`, `object_type`, and `tenant_id` on create/update, but the get/list responses only return `object_id` and `tenant_id`. Since `object_type` is a `Required` (non-Computed) attribute in the resource schema, Terraform expects the final state to exactly match the planned/config value — the provider currently overwrites it with `null` from the API response, triggering the error.

## ✨ Description of new changes

- `internal/services/onelakedataaccesssecurity/models.go`: added `reconcileMicrosoftEntraMemberObjectTypes`, which fills in a missing `object_type` for `microsoft_entra_members` entries (matched by `object_id`) using the known value from the plan (Create/Update) or the prior state (Read), after the model is populated from the API response.
- `internal/services/onelakedataaccesssecurity/resource_onelake_data_access_security.go`: call the reconciliation helper in `Create`, `Update`, and `Read`, right after `r.get(...)` and before the state is persisted.
- Also fixed a closely related bug in the same request-building code path: `requestCreateOrUpdateOneLakeDataAccessSecurity.setMembers` always sent an empty (non-nil) `fabric_item_members` array even when the config omitted it entirely. This produced the same class of "inconsistent result" error (`was null, but now cty.SetValEmpty(...)`) for any role that only configured `microsoft_entra_members` without `fabric_item_members` — which is exactly the scenario in the linked issue. Now the field is left `nil` when there are no fabric item members.
- `internal/services/onelakedataaccesssecurity/fake_test.go`: added `fakeGetDataAccessRoleFuncWithoutEntraObjectType`, a fake `GetDataAccessRole` implementation that mirrors the real API behavior by stripping `object_type` from the response.
- `internal/services/onelakedataaccesssecurity/resource_onelake_data_access_security_test.go`: added `TestUnit_OneLakeDataAccessSecurityResource_MicrosoftEntraMembers_MissingObjectType`, a regression test that reproduces the exact scenario from the issue (create with `microsoft_entra_members`, API omits `object_type`) and asserts both apply and a subsequent refresh-only plan succeed with no drift.

No schema, docs, or examples changes were needed since this is a bug fix to existing behavior (no new/changed attributes).

## ☑️ PR Checklist

- [x] Link to the issue you are addressing is included above
- [x] Ensure the PR description clearly describes the feature you're adding and any known limitations

## ☑️ Resources / Data Sources Checklist

- [x] Production quality implementation of the fix in `./internal/services/onelakedataaccesssecurity`
- [x] Unit tests added/passing (`go test ./internal/services/onelakedataaccesssecurity/...` with `FABRIC_PREVIEW=true`); verified the new regression test fails without the fix and passes with it
- [ ] Not applicable: no new resource/data-source attributes, so no examples/schema description/docs/template changes were required


KRJNK:
@microsoft-github-policy-service agree

